PR #25, opened 2021-10-20
Repo dev and CI/CD simplification

jharvey10:
We should definitely chat about this PR before I merge it in, but I think this will be a lot simpler to get started and manage the workflow of the monorepo.

I was reflecting on how many steps there were to get the repo set up and how we've all (especially me) been fiddling between npm and lerna commands. I can do better! The last thing I want is for the dev experience to be weighing us down.

### The previous setup steps to get up and running:
1. Install `nvm` to manage node versions
    - install via `curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | bash`
2. `nvm install v16` (and npm 8)
3. `nvm use 16`
4. `npm -g install lerna`
5. `npm install`
6. `lerna bootstrap`
    - ** fyi: regarding spacing - reformats package-lock.json --> https://github.com/lerna/lerna/issues/2845
7. `npm run all:build`
8. Start developing

### The new setup steps are:
1. Install `nvm` to manage node versions
    - install via `curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | bash`
2. `nvm install v16` (and npm 8)
3. `nvm use 16`
4. `npm install`
5. `npm run packages:build`
6. Start developing

### Things included in this PR:

- **Removal of lerna.** I keep running into bugs with Lerna and the way it updates package.json, package-lock.json, and repo tags/changelogs (things like tabs vs spaces, lockfile versions left un-updated, duplicate entries in changelogs, etc.). It might be because we're using npm 7/8, but considering how long 7 has been out, there's no excuse for anything other than full support at this point. Lerna hasn't had even a patch release since February of this year, and the release before that was somewhere around 17 months ago. There's 652 open issues and 63 un-merged pull requests with no maintainer attention to speak of.
- **Addition of `micromanage`.** What we need to do in terms of installing, linking, versioning, and publishing is really not that complicated, but the piece that's missing from existing solutions is the management/publishing/deployment of both npmjs packages *and* microservices that get deployed to a cloud environment. So I took a couple hours and just wrote the scripts myself and called it `micromanage`.
- **Reintroduction of npm workspaces.** Since `micromanage` is fully aware of npm workspaces, they become the de-facto way to work with packages in the repo.
- **Removal of project-level `package-lock.json` files.** This goes along with the change to reintroduce npm workspaces. Project-level package-lock files don't exist in a workspaces environment. Nor do project-level `node_modules` dirs except in the case of dependency version conflicts between packages.

### Coming soon:

- Implement the cloud deployment logic in `micromanage`
- Incorporate `micromanage link` into pre-commit hooks
  - This is needed because npm workspaces will allow you to use local packages without running `npm install` against them for the workspaces in which they are used. `micromanage link` searches through the project source `.ts` files, finds usages of local packages, and adds them to the appropriate `package.json` files in which they are used.


mattrosno (commented):
Can you add documentation to either the README or CONTRIBUTING file on how to install dependencies for the repository, or for a specific workspace? Also, guidance if it's best to step into a workspace and run commands like `npm run dev` from there, or if it's best to do that from the root repository?

The only trouble I was running into with the former setup was how you had to `npm install` from both the root level directory and within `/services/web-app`.

jdharvey-ibm:
Since we (briefly) went through this on a call and in the interest of seeing how it works tonight in the nightly build, I'll merge this in since it won't break any of the existing packages/services.

PR #27, opened 2021-10-20
GitHub Pages Scaffolding

jharvey10:
Basic setup to create a github pages site based on the docs folder.

Using the "just the docs" theme, which has some basic search functionality

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -1,0 +1,1 @@
+remote_theme: pmarsceill/just-the-docs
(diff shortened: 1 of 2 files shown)


mattrosno (approved):

mattrosno:
https://pmarsceill.github.io/just-the-docs/ looks cool and all, but might be overkill at this stage, especially if it becomes a time sink to configure and customize it. I think we can get by with repo markdown files for now.

PR #54, opened 2021-10-27
feat: add component catalog base

### LINKED ISSUE #12: Create components needed for asset catalogs

andreancardona:
Closes #12

Related to #8 

#### Changelog

**New**

- Adds in Catalog List View
- Adds in Catalog Grid View
- Breaks up each component into their own directory and stylesheet

#### Testing / reviewing

- [ ] Make sure you see _both_ view styles for the component catalog at all the breakpoints

Screen Shot 2021-11-08 at 11 30 11 AM

#### Note: 
- This does not include functionality 
- Still missing filter and search, pagination styling & functionality


mattrosno (commented):
@andreancardona can this PR close https://github.com/carbon-design-system/carbon-platform/issues/12, so we don't have to wait to merge this until the full prototype is done?

mattrosno (commented):
Next.js file naming convention is to use lowercase dashed file names like `component-catalog.js` instead of `ComponentCatalog.js`. Not a big deal right now, but let's standardize on that going forward.

I'm thinking these components can go in a `/components/catalog` directory, as the catalog components will be not specific to components... they will be used in the Elements catalog, Patterns catalog, etc. All asset catalogs.

Actually, we can split these components up to their own directories so each components can have its own CSS Modules file. That way, we can use short class names like `styles.container` and not need `styles.componentCatalogItemGridContainer`.

mattrosno (dismissed):
Looking good! Lmk if you want to pair up to do any of these recommendations.

andreancardona:
> Next.js file naming convention is to use lowercase dashed file names like `component-catalog.js` instead of `ComponentCatalog.js`. Not a big deal right now, but let's standardize on that going forward.
> 
> I'm thinking these components can go in a `/components/catalog` directory, as the catalog components will be not specific to components... they will be used in the Elements catalog, Patterns catalog, etc. All asset catalogs.
> 
> Actually, we can split these components up to their own directories so each components can have its own CSS Modules file. That way, we can use short class names like `styles.container` and not need `styles.componentCatalogItemGridContainer`.


- [ ]  Next.js file naming convention is to use lowercase dashed file names like `component-catalog.js` instead of `ComponentCatalog.js`. Not a big deal right now, but let's standardize on that going forward.

- [ ] I'm thinking these components can go in a `/components/catalog` directory, as the catalog components will be not specific to components... they will be used in the Elements catalog, Patterns catalog, etc. All asset catalogs.

- [ ]  Actually, we can split these components up to their own directories so each components can have its own CSS Modules file. That way, we can use short class names like `styles.container` and not need `styles.componentCatalogItemGridContainer`.

mattrosno (approved):

Issue #2, opened 2021-10-13
Define and document library and asset schemas


mattrosno:
### Summary

Hill 1A: _An IBM Maker [designers, developers, product managers delivering to the IBM ecosystem] can **discover and learn** about resources [standards and **components/patterns**] in the system with confidence in their completeness, who maintains them and where they're used._

Regardless of where those resources come from (which GitHub repository, GitHub public or enterprise), regardless of where those resources are currently documented (Carbon website, PAL sites), we need to fully understand each resource to properly provide a discovery experience.

That means we need to standardize what it means to be a library and an asset. We need to define what properties belong to each, and provide a mechanism for libraries and assets to describe themselves so Carbon's platform can index those resources and present them through one or many catalogs.

This can be achieved by creating a schema and documenting how to apply it, so we can later apply it to the various libraries and assets in Carbon's ecosystem.

If these are our initial entities and relationships, this effort will define which properties belong to the library and asset entities.

![image](https://user-images.githubusercontent.com/1691245/137197338-a0ffd2a5-c31d-4a37-b07f-dbc446725eeb.png)

### Job stories

When I'm searching for a solution for a user need,
I want to see fully-qualified assets (status, framework, maintainers, dates, usage, etc.)
so that I can find the asset I need.

When I'm looking for an asset,
I want to look at only assets that are stable and trustworthy
so that I don't waste time with looking for assets that are unstable or that I and my development team won't trust.

When I'm looking for an asset,
I want to find assets based on what kind of asset they are (elements, components, patterns, shared services, etc.)
so that I don't waste time looking for assets that are inappropriate for the problem at hand.

When I'm looking for an asset,
I want to find assets based on what kind of asset they are (e.g. input, navigation, information, container)
so that I don't waste time looking for assets that are appropriate for the problem at hand.

When referencing components and patterns maintained by different teams,
I want them to be consistently classified
so I don't have to remember multiple naming and classification schemes.

When I am grabbing an asset for usage,
I want to know what framework the asset is available in
so that I know if it is or isn't compatible with our product, and whether we can implement it.

When designing for a product built with a specific JavaScript framework,
I want to use design assets and guidance that has been implemented by that framework,
so there's no misunderstanding during handoff.﻿

When publishing the documentation for component I maintain,
I want to specify metadata such as its status, framework, platform, version, etc.
so my users can fully qualify the asset and determine if it fits their needs or not.

When determining to use a library or not,
I want to know if I can technically and legally use it,
[...]

PR #65, opened 2021-11-02
docs: add library schema docs

### LINKED ISSUE #4: Document the schema structure and values
We'll need to document the schema(s) in terms of what they are, how to apply them, and the available values per property.

This could probably live in the top-level `/docs` directory in this repository for starters?

In the future, we might want to put the documentation into a package that we can independently version. Should we eventually ship it as a code package, that means we could deliver a script used by each repository to call in a pre-commit check where the script finds schemas and verifies that each Carbon config is valid.

A good example is the new GitHub issue templates. It states, "issue forms follow a schema, here's an example, here are the available keys, types, and value values."

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema

mattrosno:
Closes #4

Adds documentation for how to add library and asset metadata files. Additional properties will be added in #6.

#### Changed

* Simplified the response of the image optimization
* Updated the prototype to use `thumbnailPath` instead of `thumbnail`

#### Review

Here's the markdown documentation.

https://github.com/carbon-design-system/carbon-platform/blob/b2e37d07d901f8e3f084c71a80ee484e9ca7f031/docs/resource-schemas.md


mattrosno:
@andreancardona can you review this? In this first round of applying the schema to known libraries/assets, I'm thinking we'll just include the properties that I've documented in this PR: https://github.com/carbon-design-system/carbon-platform/blob/6607ad44c42b0b4b4f76152acff4a57e6e62bedf/docs/resource-schemas.md

Then after this feels good, via #6 we can add more properties to the schema, to the metadata files, then in the prototype.

jdharvey-ibm:
@mattrosno Should we add a card to the backlog to turn the schemas into json/yaml schema files? If we have those on GitHub, they could be referenced by other projects to get code completion and property validation in real-time as they're being edited.

mattrosno:
@jdharvey-ibm I don't follow your question. The application of the schema will take place in various repos in yml metadata files like this example: https://github.com/mattrosno/carbon-next/blob/e2f8bff6883f69a01feeb134b4b8a66414b5b07e/packages/react/src/components/Accordion/carbon-asset.yml

Our prototype is currently reading that YML data and parsing it into JSON.

mattrosno:
@jdharvey-ibm bottom two issues in this epic will apply the schema for prototyping purposes.

https://github.com/carbon-design-system/carbon-platform/issues/2

jdharvey-ibm:
@mattrosno Ah sorry, I'll try to clarify. I mean an actual formal json schema file that defines all possible fields that can exist in a `carbon-asset.yml` file or other file governed by a formal schema. This would be something like a `carbon-asset.schema.json` file that defines what fields are valid, what types they are, and where they can be used in one of the `carbon-asset.yml` files that an end user would place in their repo.

To use a schema, you include a line in your file to reference it, and from there, you get validation and auto-completion. Here's some pictures demoing me pulling in the example schema over at [json-schema.org](https://json-schema.org/learn/examples/address.schema.json)

image
image
image

mattrosno:
@jdharvey-ibm got it - I'll document as a separate issue for schema tooling.

@andreancardona I added more documentation to describe value values: https://github.com/carbon-design-system/carbon-platform/blob/cdb912fa70c3cbe6dce4ddbef76bd7bb425e1b53/docs/resource-schemas.md

mattrosno:
Markdown file has been updated: https://github.com/carbon-design-system/carbon-platform/blob/b2e37d07d901f8e3f084c71a80ee484e9ca7f031/docs/resource-schemas.md

andreancardona (approved):

PR #71, opened 2021-11-09
style: add import sorting

### LINKED ISSUE #64: Add import sorting
Matt's currently using this VS Code extension: https://marketplace.visualstudio.com/items?itemName=amatiasq.sort-imports

From Joe

>Regarding import sorting, I have had really good luck with the eslint simple-import-sort module. It's opinionated and tends to do a pretty good job with separating out local vs vendor deps as well as alphabetizing them.

It'd be good to add import sorting to our linting and editor configurations.

jharvey10:
- also fix double star in copyright headers

Closes #64 

#### Testing / reviewing

Note that this causes the timing issue with the style prefixes as reported over in https://github.com/carbon-design-system/carbon/discussions/9723 to manifest in dev builds in addition to production builds. This can be worked around by running the fix script prior to starting the dev server.

### DIFF
--- a/services/web-app/pages/_app.js
+++ b/services/web-app/pages/_app.js
@@ -4,13 +4,20 @@
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
  */
+/* eslint-disable simple-import-sort/imports */
+/*
+ * TODO: fix import sorting once Carbon prefix bug is resolved.
+ * Layout must be imported before carbon-components-react
+ * Details: https://github.com/carbon-design-system/carbon/discussions/9723
+ */
 import '@/styles/styles.scss'
+
+import Layout, { LayoutProvider } from '@/layouts/layout'
 
 import { unstable_FeatureFlags as FeatureFlags } from 'carbon-components-react'
 import { DefaultSeo } from 'next-seo'
 
 import defaultSeo from '@/config/seo.json'
-import Layout, { LayoutProvider } from '@/layouts/layout'
 
 function App({ Component, pageProps }) {
   return (
--- a/services/web-app/pages/_app.js
+++ b/services/web-app/pages/_app.js
@@ -1,17 +1,16 @@
-/**
+/*
  * Copyright IBM Corp. 2021, 2021
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
  */
-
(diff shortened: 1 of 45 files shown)


mattrosno (dismissed):
That v11 script that we run just affects node_modules, right? How would import sorting affect the files in node_modules that we are overriding in our builds?

Do we need to document any VS Code extensions or settings that we should all use with this?

services/web-app/pages/_app.js
```diff
@@ -1,17 +1,16 @@
...
 import '@/styles/styles.scss'
 
-import Layout, { LayoutProvider } from '@/layouts/layout'
-
-import { DefaultSeo } from 'next-seo'
 import { unstable_FeatureFlags as FeatureFlags } from 'carbon-components-react'
+import { DefaultSeo } from 'next-seo'
+
 import defaultSeo from '@/config/seo.json'
+import Layout, { LayoutProvider } from '@/layouts/layout'
```
jdharvey-ibm, line 13: @mattrosno You're correct that the script only affects the node modules, but one or more of the imports have side effects (which is probably the root of the problem). Changing the `layout` to be imported _after_ carbon components react is what causes the issue to manifest. If I reverse those two lines here, I don't see the problem.

jdharvey-ibm:
> That v11 script that we run just affects node_modules, right? How would import sorting affect the files in node_modules that we are overriding in our builds?
> 
> Do we need to document any VS Code extensions or settings that we should all use with this?

@mattrosno The current dev env instructions should cover all of the necessary setup. Since it's eslint-based, as long as your IDE has eslint set up (and you don't have any other conflicting extensions installed that are trying to modify the import order), it should work without any additional setup.

services/web-app/pages/_app.js
mattrosno: Can we `// eslint-disable-next-line simple-import-sort/imports` in that file with side effects until that v11 bug is fixed? (The comment has to be at the start of the file's imports.)

Docs: https://github.com/lydell/eslint-plugin-simple-import-sort/blob/main/examples/ignore.js#L17

francinelucca (dismissed):

services/web-app/pages/_app.js
jdharvey-ibm: Yeah, I'll add that to this PR.

mattrosno (approved):
Installed, ran full build, started web app and styling looks good.

PR #76, opened 2021-11-12
feat: add sponsor team to each library

mattrosno:
Related #6

Libraries can have a sponsor team. This is different than maintainers, which are individuals. It's likely that the majority of maintainers will belong to the sponsor team for any given library.

By specifying a sponsor team, we can show that in the catalogs that there is a group of IBMers, a team, committed to maintaining the library. By "sponsoring", we mean committing their time to act as the primary group of maintainers.

Each sponsor team will have a name and a "mark" (an icon). Instead of adding the sponsor team to each schema, because this is something that will likely be stored in our database, we can simply specify sponsor teams in the libraries allowlist.

See comment in `/data/teams.js` for how we can use this in the catalogs.

#### Changelog

**New**

- Adds initial data structure for sponsor teams that will grow as we index more libraries

### DIFF
--- a/services/web-app/data/libraries.js
+++ b/services/web-app/data/libraries.js
@@ -14,151 +14,176 @@ const libraryAllowList = {
     host: 'github.ibm.com',
     org: 'matt-rosno',
     repo: 'carbon',
-    path: '/packages/carbon-react'
+    path: '/packages/carbon-react',
+    sponsor: 'carbon'
   },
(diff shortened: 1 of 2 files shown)


mattrosno:
@andreancardona let me know if you have questions with this, as the starting point for us to show sponsor teams in the catalog items.

francinelucca (approved):
LGTM

PR #77, opened 2021-11-12
feat: add private property to resource schemas

mattrosno:
Related #6 

Adds an option to exclude libraries and assets for the catalogs. By default, all indexed libraries and assets will be included in the catalogs. This follows how npm uses the `private` key. See: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#private

In the future, we'll have some "reviewed" or "verified" property that will be used in the catalog's default filters. To counter-balance `private` which defaults to showing everything, `reviewed` will default to not reviewed so the resource can go through the review process.

#### Changelog

**New**

- `private` key in the schemas, defaulting to false
- Library and asset catalog pages filter out private items

#### Testing / reviewing

See docs file that explains the `private` key.

### DIFF
--- a/docs/resource-schemas.md
+++ b/docs/resource-schemas.md
@@ -26,12 +26,13 @@ description: React implementation of Carbon Components
 
 ### Library keys
 
-| Key               | Description                                                                                                                    | Required | Type   | Default         | Valid values |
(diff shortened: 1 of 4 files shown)


mattrosno:
We can ignore the failed checks since @jdharvey-ibm changed the action configs since this PR was opened?

jharvey10 (approved):

PR #78, opened 2021-11-12
feat: add asset inheritance

### LINKED ISSUE #7: Schema library and asset inheritance
To do: document the component framework example, multi-framework monorepo example, and rollup example with supporting job stories.

mattrosno:
Related: #6 
Closes #7 

Adds an `id` to each asset and defines an inheritance relationship between assets so properties like name, description, thumbnail don't have to be duplicated. This also allows single source of truth for usage guidance, design specifications, and styling to be used in multiple implementations.

In the future, when we add additional content that can be inherited like usage guidance and design specs. I expect that would get documented in the `properties` property of `inherits`.

#### Changelog

**New**

- Library `description`, `license`, `package`, `version` from `package.json`
- Asset `id`
- Asset `inherits`

#### Testing / reviewing

See docs and example for inheritance here: https://github.com/carbon-design-system/carbon-platform/blob/cd67cf8458479b401741ab29c824ddd70881b257/docs/resource-schemas.md

#### To do

- [x] Placeholder for inherited image
- [x] Link to inherited asset


mattrosno:
@jdharvey-ibm I documented your suggestions here: https://github.com/carbon-design-system/carbon-platform/issues/126

IMO, we should consider those later and first see what we learn when applying the approach in this PR for prototyping purposes.

francinelucca (dismissed):

PR #127, opened 2021-11-16
fix(web-app): bx vs cds prefix issue

jharvey10:
DO NOT MERGE this PR until the fix for https://github.com/carbon-design-system/carbon/pull/10128 is available in a release and has been picked up on this branch.

- Remove carbon prefix workaround script
- Remove FeatureFlags from _app.js

PR #129, opened 2021-11-17
feat: asset catalog protoype

### LINKED ISSUE #31: Asset catalog item components for prototype
Presentational components for the catalog items as a grid view:

image

And list view:

image

These can be static components that we populate with data when the data is ready. As the deliverable is a prototype for user testing and design iteration, no need for pixel perfection, prop types documentation, etc.

andreancardona:
Closes #31 

## Styling todos

For both List & Grid view:
- [x] Add fill for content switcher svgs @mattrosno 
- [x] Center `sponsor tags` @mattrosno 
- [x] Add launch svgs on hover @mattrosno 
- [x] Thumbnail component via component name matching @mattrosno 
- [x] Add all framework and team icons @mattrosno 

## @mattrosno review

- [x] `CatalogGrid`
- [x] `CatalogIndexPage`
- [x] `CatalogList`
- [x] `CatalogSearch`
- [x] `CatalogSort`
- [x] Sonarlint errors
- [x] Icons
- [x] Breakpoints
- [x] Color tokens
- [x] Spacing tokens


mattrosno:
@andreancardona to (1) fix lint errors and (2) merge in `main` branch then I'll review.

services/web-app/components/catalog-grid/catalog-grid-item.js
```diff
@@ -0,0 +1,96 @@
...
+                    : asset.content.framework === 'vanilla'
+                      ? (
+                  <Svg24Vanilla />
+                        )
+                      : asset.content.framework === 'vue'
+                        ? (
+                  <Svg24Vue />
+                          )
+                        : null}
+              </div>
```
jdharvey-ibm, line 87: Hmm, could this instead be offloaded to its own component? Maybe something that has a map of framework names to icons, like:
```jsx
const frameworkToIconMap = {
  angular: Svg24Angular,
  react: Svg24React,
  etc...
}

const FrameworkIcon = ({name}) => {
  const ActualIcon = frameworkToIconMap[name]

  return <ActualIcon />
}
```

jdharvey-ibm: Something like that might help out with the sponsor tags above as well.

andreancardona: Yep! I was trying something like this, and was running into an issue. So this was my temporary solution. Mind if wee look into it tomorrow together?

mattrosno: I'm making a pass on this tomorrow. I added sonarlint errors to the list in the PR description above.

mattrosno:
@andreancardona @jdharvey-ibm I'm addressing the above todos in https://github.com/carbon-design-system/carbon-platform/pull/134

mattrosno (changes requested):
Looks like some bugs only when doing a production build.

- [x] The inline notification has a "title" in there
- [x] Search bar is the wrong height
- [x] "Sort by:" is misaligned, and the "A-Z" isn't right aligned

image

mattrosno:
I logged the issues here: https://github.com/carbon-design-system/carbon/discussions/9723

@andreancardona let's wait to merge this PR until we can pull in a newer version of `@carbon/react` that contains this fix: https://github.com/carbon-design-system/carbon/pull/10128

mattrosno (changes requested):
From @jeanservaas:

List view, to prevent long asset description line lengths:

- [x] `lg` breakpoint, content should be 7 columns
- [x] `xlg` breakpoint and bigger, content should be 6 columns

mattrosno (changes requested):
We have a couple "off by 1px" breakpoint bugs.

- [x] At exactly the `lg` breakpoint, the license should be inline with stable
- [x] At exactly the `md` breakpoint, the catalog item should not be hanging into the grid gutter

image

image

mattrosno:
Tracking the `@carbon/react` issues here: https://github.com/carbon-design-system/carbon-platform/pull/127#issuecomment-977929202

mattrosno (approved):

PR #139, opened 2021-11-30
chore: update stylelint rules and update hover borders on catalog

mattrosno:
#### Changelog

**New**

- Adds `title` attributes to sponsor icon, framework icon, and status icon. We can't use a tooltip here because it's containing anchor is interactive; the icons aren't interactive.
- New `framework` data object so we can use display names.
- Adds contextual subtle select border colors when catalog items are hovered and focused with transitions.

**Changed**

- Stylelint rules so we can have one level nesting and so mixins and pseudo state declaration blocks aren't counted
- Refactored component styles with nesting so the stylesheets are more legible

**Removed**

- N/A

#### Testing / reviewing

Component catalog should look exactly as it did before with darker borders on catalog item hover/focus.

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -4,6 +4,13 @@ module.exports = {
   reportNeedlessDisables: true,
   reportInvalidScopeDisables: true,
   rules: {
-    'selector-pseudo-class-no-unknown': null
+    'selector-pseudo-class-no-unknown': null,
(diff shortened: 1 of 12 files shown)


francinelucca (approved):
Ran changes locally, LGTM and can perceive darker borders

jharvey10 (approved):

PR #149, opened 2021-12-02
feat: multiselect filter component

### LINKED ISSUE #32: Asset catalog filtering components for prototype
These designs will be changing, but we need an initial way to interact with catalog data.

Screen Shot 2021-12-01 at 11 53 25 PM

Like the rest of the prototype, these presentational components don't need to be production-ready; just enough for us to user test.

### LINKED ISSUE #158: Refactor: cleanup filter tagging logic
### Summary

Related #149 

Right now the logic for the selected filters in the overflow state is very basic. Each breakpoint currently has it's own logic. 
There must be a better way to accomplish this. 

## Refactoring

- [x] Refactor: `catalog-multiselect-filter.js` lines: `145-230`

Screen Shot 2021-12-09 at 10 14 48 AM

andreancardona:
Closes #32 & Closes #158

Created a separate issue to refactor the overflow filter tag logic at different breakpoints: https://github.com/carbon-design-system/carbon-platform/issues/158

#### Changelog

**New**

- Adds base for filter component

### Design / Visual:
- [x] Align Multiselect Dropdown to full Grid width (in XL, L & MD breakpoints)
Screen Shot 2021-12-15 at 2 32 46 AM

- [x] `overflowMenu` tag needs to flip left when triggered

- [x] `overflowMenu` tag needs to show filter category -- ex: `Sponsor`

- [x]  Mobile: need to add filter tag count - there currently is no count on the icon

- [x]  Mobile: need to fix search bar and filter icon widths to match specs below
Screen Shot 2021-12-15 at 2 39 24 AM

- [x]  Mobile: In mobile when the multiselect dropdown is open -- icon needs to transform to a `close` icon
Screen Shot 2021-12-15 at 2 41 03 AM

### Bugs: 
- [x] Fix bug --> Multiselect Dropdown state remains false after the `overflowMenu` tag is triggered 

- [x] `overflowMenu` needs to remain relative to the overflow menu tag
Screen Shot 2021-12-15 at 2 37 33 AM


andreancardona:
@mattrosno making this PR _mainly_ to start getting feedback on how we can improve / refactor earlier on in the process. Would love your thoughts. Will keep working on this in the mean time off of this branch.

services/web-app/components/catalog-search/catalog-search.module.scss
```diff
@@ -19,3 +19,98 @@
 .container :global(.cds--label) {
   display: none;
 }
+
+.multiselect-input {
```
mattrosno, line 23: We updated stylelint rules to allow nesting, so you can now do:

```scss
.foo {
  ...

  &:focus,
  &:hover {
    ...
  }
}
```

services/web-app/components/catalog-search/catalog-search.module.scss
```diff
@@ -19,3 +19,98 @@
...
+  width: 100%;
+  padding: $spacing-05;
+  border-left: 1px solid $border-subtle;
+
+  @include breakpoint(md) {
+    width: 50%;
+  }
+
+  @include breakpoint(lg) {
+    width: 33.33%;
```
mattrosno, line 83: Is there a Sass function like `percentage(1 / 3)` that we can use for more decimal precision without typing out `33.333333%`?

services/web-app/components/catalog-search/catalog-search.js
```diff
@@ -40,15 +151,46 @@ const CatalogSearch = ({ onSearch }) => {
...
-            <MultiSelect
-              id="catalog-filter"
-              label="Filters"
-              items={items}
-              itemToString={(item) => (item ? item.text : '')}
-              initialSelectedItems={[items[0], items[1]]}
-              size="lg"
-            />
+          <Column sm={1} md={4} lg={4}>
+            <button
```
mattrosno, line 155: What if all this logic was in a new `MultiSelect` component so its `handleSelect()`, `handleRemoveItem()`, `handleRemoveAll()` would be local to that component's source files and not part of the `CatalogSearch` component?

I don't know a good name... `MultiSelectFilter` or something?

mattrosno (commented):
I don't know the "Carbon" way to do this... but is there any way to use a Carbon-provided primitive to control if the multi-select is shown or not?

E.g., instead of `const [triggerMultiselect, setTriggerMultiselect] = useState(false)`... we use something so it's impossible for Carbon's `Dropdown` and this custom component to be open at the same time?

I think that means using the `ListBox` component.

https://github.com/carbon-design-system/carbon/tree/main/packages/react/src/components/ListBox

services/web-app/components/utils/useOutsideClick.js
```diff
@@ -0,0 +1,25 @@
...
+import { useEffect } from 'react'
+
+const useOutsideClick = (ref, callback) => {
+  const handleClick = (e) => {
+    if (ref.current && !ref.current.contains(e.target)) {
+      callback()
+    }
+  }
+
+  useEffect(() => {
```
mattrosno: Should this `useEffect` have an empty dependency array `[]` to prevent it from running too much? If not, let's add a comment to clarify this as it's expected that `useEffect`s have stated dependencies.

services/web-app/components/utils/useOutsideClick.js
```diff
@@ -0,0 +1,25 @@
+/*
```
mattrosno: Can this go in `/services/web-app/utils` directory, renamed to `use-outside-click.js` to follow naming conventions that you'll see once you pull in latest from main branch?

services/web-app/components/catalog-search/catalog-search.js
```diff
@@ -4,14 +4,15 @@
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
  */
-import { Column, Grid, Layer, MultiSelect, Search, Theme } from '@carbon/react'
+import { Column, Grid, Layer, Search, Theme } from '@carbon/react'
 import { useState } from 'react'
 
 import { useMediaQueryContext } from '@/contexts/media-query'
 
+import CatalogMultiselectFilter from '../catalog-multislect-filter/catalog-multislect-filter'
```
mattrosno, line 12: A better import would be with the Next alias by using:

```jsx
import CatalogMultiselectFilter from '@/components/catalog-multislect-filter'

```

services/web-app/components/catalog-search/catalog-search.js
```diff
@@ -40,16 +35,7 @@ const CatalogSearch = ({ onSearch }) => {
...
-            <MultiSelect
-              id="catalog-filter"
-              label="Filters"
-              items={items}
-              itemToString={(item) => (item ? item.text : '')}
-              initialSelectedItems={[items[0], items[1]]}
-              size="lg"
-            />
-          </Column>
+          <CatalogMultiselectFilter onSelect={onSelect} />
```
mattrosno, line 38: I'd leave the `<Column>` here. Ideally, a component simply fills its parent's container. Also, it's not like this component will be used elsewhere, but if it will be, it would be nice if it didn't have it's own layout logic. That should belong to it's parent.

services/web-app/components/catalog-multislect-filter/catalog-multislect-filter.js
```diff
@@ -0,0 +1,177 @@
...
+ * Copyright IBM Corp. 2021, 2021
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import { Column, Tag } from '@carbon/react'
+import { ChevronDown, ChevronUp, OverflowMenuHorizontal } from '@carbon/react/icons'
+import { useRef, useState } from 'react'
+
+import { useOutsideClick } from '@/utils/use-outside-click'
```
mattrosno, line 11: Not sure if custom outside click events will be necessary if you use ListBox components to make this custom dropdown filter.

https://github.com/carbon-design-system/carbon/tree/main/packages/react/src/components/ListBox

services/web-app/components/catalog-multislect-filter/catalog-multiselect-filter.module.scss
```diff
@@ -0,0 +1,120 @@
...
+.container {
+  width: 100%;
+  height: $spacing-09;
+  border: none;
+  border-left: 1px solid $border-subtle;
+  background: $background;
+  box-shadow: inset 0 -1px 0 $border-strong;
+  cursor: pointer;
+
+  &:focus {
```
mattrosno, line 17: Notice how the component jumps a couple pixels when you click it? That's because you're adding a border that didn't previously exist. Look at how we're doing the focus state in the catalog item. You can use outline with a negative offset. Outline doesn't take up space in the box model, and the negative offset moves it inside the component. Also, we have a token for focus color.

```scss
  &:active {
    outline: 2px solid $focus;
    outline-offset: -2px;
  }
```

andreancardona:
@mattrosno thank you for the feedback! :)

mattrosno (approved):

PR #165, opened 2021-12-21
feat: page header component

### LINKED ISSUE #167: PageHeader Component
### Summary

Create a PageHeader Componoent that can be reused throughout all of the catalog pages that match the following spec. 

- PageHeader component should have a `title` and `pictogram` prop - since the titles and pictograms will change for each page

Screen Shot 2021-12-22 at 12 01 27 PM

andreancardona:
Closes #167 

## Changed: 

- Added Page Header to components page with a `title` and `pictogram` prop
- Added a `PageHeaderPictogram` component to keep track of the different pictograms for each page 
- Added Carbon Pictograms package for component page at `service/web-app` level
- Added line clamp mixin for really long page titles: 
Screen Shot 2021-12-21 at 3 24 24 PM

- Removed content { spacing } in `pages.modules.scss` since it was adding an unnecessary margin as seen below:
Screen Shot 2021-12-21 at 3 31 28 PM


package.json
```diff
@@ -56,5 +56,8 @@
     "stylelint": "^13.13.1",
     "stylelint-config-carbon": "^1.7.0",
     "ts-jest": "^27.0.7"
+  },
+  "dependencies": {
+    "@carbon/pictograms-react": "^11.20.0"
```
mattrosno, line 61: I see you installed this at the workspace level, not the service level. Can you uninstall it from here and do:

`npm --workspace services/web-app install @carbon/pictograms-react@latest`

See docs: https://github.com/carbon-design-system/carbon-platform/blob/main/docs/day-to-day-development.md#adding-new-project-dependencies

services/web-app/pages/assets/components.js
```diff
@@ -26,6 +30,14 @@ const Components = ({ librariesData }) => {
   return (
     <>
       <NextSeo {...seo} />
+      <Grid className={styles.container}>
```
mattrosno, line 33: We need this page header on multiple pages. Can you move this to a new `PageHeader` component, that we use in all ~6 catalog pages?

I'd expect the props for the PageHeader component would be `heading` as a string, and `icon` which we'd pass in a pictogram so each page could use a different pictogram.

services/web-app/pages/assets/components.module.scss
```diff
@@ -0,0 +1,63 @@
+//
```
mattrosno, line 1: The design spec specifies truncation after two lines. Can you make sure the page header holds up with really long headings? I think you'll want to use the `line-clamp` mixin.

services/web-app/pages/assets/components.js
mattrosno: Can you add the page header to all catalog pages?

mattrosno (changes requested):

mattrosno:
image

@andreancardona I'd only use `- [ ]` checkbox markdown in PR descriptions if there are outstanding tasks.

mattrosno (changes requested):
I don't see the need for `PageHeaderPictogram` component and its related data file. Would this get used in other places than the `PageHeader` component?

Can the `PageHeader` component take a `pictogram` prop?

Each page that uses the `PageHeader` imports the pictogram.

`import { TextInput } from '@carbon/pictograms-react'`

Then when you use the component:

`<PageHeader title="My page" pictogram={TextInput} />`

And then in the `PageHeader` component you should be able to:

```jsx
const MyComponent = ({pictogram: Pictogram}) => {
  return (
    <>
       {Pictogram && <Pictogram />}
    </>
  )
}
```

So you can pass in and render a pictogram.

services/web-app/pages/index.js
```diff
@@ -4,9 +4,11 @@
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
  */
+import { TextInput } from '@carbon/react'
```
mattrosno, line 7: Wrong import. This is not a pictogram.

services/web-app/pages/standards.js
```diff
@@ -4,9 +4,11 @@
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
  */
+import { TextInput } from '@carbon/react'
```
mattrosno, line 7: Wrong import.

services/web-app/components/page-header/page-header.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import { Column, Grid } from '@carbon/react'
+
+import styles from './page-header.module.scss'
+
+const PageHeader = ({ title, pictogram: Pictogram }) => {
+  console.log(Pictogram, 'hi')
```
mattrosno, line 12: Remove console.log.

services/web-app/components/page-header/page-header.js
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <Grid className={styles.container}>
+      <Column className={styles.section} sm={4} md={5} lg={8}>
+        <div className={styles.title}>{title}</div>
```
mattrosno, line 16: `<h1>` would be more fitting.

services/web-app/components/page-header/page-header.js
```diff
@@ -0,0 +1,27 @@
...
+      </Column>
+      {Pictogram && (
+        <Column className={styles.pictogram} sm={0} md={5} lg={8}>
```
mattrosno, line 19: I don't follow how these sm / md / lg add up with the first column.

mattrosno (changes requested):
`/data/pictogram.js` needs deleted.

I think `page-header.js` can be cleaned up with this so the columns add up, so we're not overriding `Grid` with flexbox layout, so layout styling is isolated to "column" class names.

```jsx
import { Column, Grid } from '@carbon/react'
import clsx from 'clsx'

import styles from './page-header.module.scss'

const PageHeader = ({ title, pictogram: Pictogram }) => {
  return (
    <Grid className={styles.container}>
      <Column className={styles.column} sm={4} md={6} lg={10}>
        <h1 className={styles.title}>{title}</h1>
      </Column>
      {Pictogram && (
        <Column className={clsx(styles.column, styles.columnLast)} sm={0} md={2} lg={2}>
          <Pictogram className={styles.pictogram} />
        </Column>
      )}
    </Grid>
  )
}

export default PageHeader
```

```scss
$page-header-height: $spacing-13 + $spacing-12 + $spacing-05;
$background-header: #d0e2ff;

.container {
  height: $page-header-height;
  background: $background-header;
  @include breakpoint-down(md) {
    padding: 0 $spacing-05;
  }

  &::before {
    position: absolute;
    z-index: 0;
    left: 0;
    width: 100vw;
    height: $page-header-height;
    background: $background-header;
    content: '';
  }
}

.column {
  position: relative;
  display: flex;
  height: 100%;
  align-items: flex-end;

  &-last {
    justify-content: flex-end;
    @include breakpoint-down(md) {
      display: none;
    }
  }
}

.title {
  margin-bottom: $spacing-07;
  color: $text-primary;
  @include type-style('display-01');
  @include line-clamp(2);
  @include breakpoint-down(md) {
    @include type-style('quotation-02');

    font-family: font-family('sans');
  }
}

.pictogram {
  margin-bottom: $spacing-08;
}
```

mattrosno (approved):

PR #166, opened 2021-12-21
feat: apply filter to lists

mattrosno:
Filter the catalog.

#### Changelog

**New**

- The default filter for all catalogs is stable status.
- Components catalog has a different default filter, as carbon sponsor, react framework, stable status.

**Changed**

- Fixed a bug when searching and the asset doesn't have a name or description.

**Removed**

- N/A

#### Testing / reviewing

Go to a URL like this to see a non-default filter, or, interact with the filter:

http://localhost:3000/assets/components?filter=%7B%22sponsor%22%3A%5B%22ibm-dotcom%22%5D%2C%22framework%22%3A%5B%22react%22%5D%2C%22status%22%3A%5B%22stable%22%5D%7D&view=grid&sort=status&q=card

Which renders:
image

### DIFF
--- a/services/web-app/components/catalog/catalog.js
+++ b/services/web-app/components/catalog/catalog.js
@@ -19,7 +19,19 @@ import { queryTypes, useQueryState } from '@/utils/use-query-state'
 
 import styles from './catalog.module.scss'
 
-function Catalog({ data, type = 'component' }) {
+const assetIsInFilter = (asset, filter) => {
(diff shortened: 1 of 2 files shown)


francinelucca (approved):
Tested with url-supplied filters and interacting with filter directly, LGTM!
Note: when using url-supplied filters on dev mode the page keeps "loading" (showing the spinner on browser tab, the page loads fine) indefinitely, but couldn't replicate when running a build so I think we should be fine.

PR #180, opened 2022-01-05
fix: use explicit key on catalogItem, prevent react unstable comp state

### LINKED ISSUE #178: Thumbnail images are wrong in the prototype
### Browser

_No response_

### Description

For some reason, the thumbnail images are wrong. As a test, setting an id on the image:

image

But then sometimes it's clearly the wrong image. The id and alt attributes are correct, but the image that it loads is wrong.

image

### Steps to reproduce

See https://carbon-platform-playground.wdc1a.ciocloud.nonprod.intranet.ibm.com/assets/components

francinelucca:
Closes #178 

Used explicit key in map for catalogItem component to prevent unstable react state, see: https://reactjs.org/docs/lists-and-keys.html#keys

#### Changelog

**Changed**

- services/web-app/components/catalog-list/catalog-list.js: include asset slug in catalog key

#### Testing / reviewing

Pull changes and verify correct app performance and thumbnail placement

### DIFF
--- a/services/web-app/components/catalog-list/catalog-list.js
+++ b/services/web-app/components/catalog-list/catalog-list.js
@@ -8,6 +8,7 @@ import { Column, Grid } from '@carbon/react'
 import clsx from 'clsx'
 
 import CatalogItem from '@/components/catalog-item'
+import { getSlug } from '@/utils/slug'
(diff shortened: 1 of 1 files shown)


mattrosno (approved):
LGTM!

PR #183, opened 2022-01-06
chore: index charts

mattrosno:
Indexes the 5 Carbon Charts libraries.

No libraries are private, as the 4 framework libraries can inherit the vanilla JS library.

Adds thumbnails for charts that match chart ids.


francinelucca (approved):
LGTM!
Pulled changes and can appreciate new charts in all frameworks: vanilla, react, angular, svelte and vue
Note: no framework icon is rendered for design-only components, assuming this is intended
image

mattrosno:
> Note: no framework icon is rendered for design-only components, assuming this is intended

@francinelucca great catch. I'll take note of this in that InVision Freehand for us to look into.

PR #189, opened 2022-01-12
feat: added breadcrumb component and styles

### LINKED ISSUE #188: Component: Breadcrumb Nav
Create a breadcrumb navigation that follows the sketched design: 
Screen Shot 2022-01-12 at 8 15 46 AM

**Note: sketch file is incorrect and the breadcrumb should only include: library of origin / id -> asset detail

andreancardona:
Closes #188 

**Note**: sketch file is incorrect and the breadcrumb should only include: library of origin / id -> asset detail

This PR adds in the breadcrumb component. Here is the sketch file:
https://ibm.ent.box.com/file/888571534535


services/web-app/pages/pages.module.scss
```diff
@@ -5,10 +5,6 @@
 // LICENSE file in the root directory of this source tree.
 //
 
-.content {
-  margin: $spacing-07 0;
-}
```
andreancardona, line 10: @mattrosno I did a visual test to see if removing this would affect any styles. Let me know if I'm missing something and we need to keep this

mattrosno: I'd search to see where that classname is all being used. I'm seeing it being used in quite a few pages.
image

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js
```diff
@@ -51,14 +53,28 @@ const Asset = ({ libraryData }) => {
   const [assetData] = libraryData.assets
   const { name, description, inherits: inheritsData, thumbnailData: imageData } = assetData.content
 
+  const id = libraryData.content.id
+
+  const libraryId = id
```
Comment (mattrosno) [inline services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:58]: Have you found `library.content.id` to not be slugified? If that's the case, see `getSlug()` in slug.js utils. Pass in content and it'll slugify id, with name as a fallback.

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js
```diff
@@ -51,14 +53,28 @@ const Asset = ({ libraryData }) => {
...
+    .replace(/\b(([a-zÁ-ú]){3,})/g, (w) => w.charAt(0).toUpperCase() + w.slice(1))
+    .replace(/-/g, ' ')
+
   const seo = {
     title: name,
-    description
+    description,
+    id: id,
```
Comment (mattrosno) [inline services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:65]: The `seo` object is used here to pass props into the `NextSeo` component. I don't think we want to add additional props in here like `id` and `libraryId` because that could be used in unexpected ways in the `NextSeo` component?

https://github.com/garmeeh/next-seo

services/web-app/components/breadcrumb-nav/breadcrumb-nav.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+import { Breadcrumb } from '@carbon/react'
+import clsx from 'clsx'
+import Link from 'next/link'
+
+import styles from './breadcrumb-nav.module.scss'
+
+const BreadcrumbNav = ({ libraryId, contentId, title, className }) => {
+  return (
+    <Breadcrumb className={clsx(styles.container, className)}>
```
mattrosno, line 15: Do we want to use Carbon's `Breadcrumb` component but not also the `BreadcrumbItem` components that are meant to go inside of it? If not using `BreadcrumbItem`, what purpose does the `Breadcrumb` serve?

mattrosno (commented):
A couple thoughts on this one.

(1) Should we be partially using the Carbon breadcrumb? One option is to also use `BreadcrumbItem` components and override for new lines and the arrow characters. Other option would be to go completely custom and not use `Breadcrumb`.

(2) Is there any way that the new `BreadcrumbNav` component can be less opinionated about libraries and assets, so it can be used in other parts of the web app as well? If doing so, there could be an `items` prop that looks something like this:

```jsx
const items = [
  {
    name: 'Carbon React',
    path: '/assets/carbon-react'
  },
  {
    name: 'Accordion',
    isCurrent: true
  }
]
```

So then the breadcrumb could take any number of items, that could have any name with any href, and one of them could be "current" so it doesn't have an underline.

Lmk if you want to talk through this approach tm!

mattrosno (approved):

PR #190, opened 2022-01-12
fix: ending / when comparing filepath to avoid picking up similar libs

francinelucca:
Find carbon-asset.yml files inside a specific library by delimitating the library name end with a /, so if the librabry name is 'carbon-cli' , we match files starting like carbon-cli/*/carbon-asset.yml and ignore files starting like carbon-cli-reporter/*/carbon-asset.yml

#### Changelog

**Changed**

- services/web-app/lib/github.js: compare with ending /

#### Testing / reviewing

test application functionality and that under functions only one reporter is rendered


mattrosno (approved):
LGTM - good catch!

PR #191, opened 2022-01-14
feat(messaging): implement messaging api

### LINKED ISSUE #80: Authentication and authorization
As a user of the carbon platform, I want to be able to log in with my IBMid so that I can see non-public, protected content that is relevant to me.

- [x] #118
- [x] #119
- [x] #120
- [x] #121
- [x] #122
- [x] #123
- [x] #125
- [x] #212

jharvey10:
- Assorted repo config and CI updates
- Update run-mode to switch to DEV and PROD as only modes

BREAKING CHANGE:

- run-mode API has changed.
- TEST is no longer a mode
- PRODUCTION is now called PROD


package.json
```diff
@@ -38,6 +38,8 @@
     "@commitlint/config-conventional": "^16.0.0",
     "@types/jest": "^27.4.0",
     "@typescript-eslint/eslint-plugin": "^5.9.0",
+    "commander": "^8.3.0",
+    "dotenv": "^10.0.0",
```
francinelucca, line 42: are we using this package anywhere?

packages/api/src/main/messaging/MessagingClient.ts
```diff
@@ -0,0 +1,184 @@
...
+  private connection?: amqp.Connection
+  private channel?: amqp.ConfirmChannel
+  private replyQueue?: amqp.Replies.AssertQueue
+  private replyCallbacks: ReplyCallbacks
+
+  private constructor() {
+    this.replyCallbacks = new Map()
+  }
+
+  public async connect(retry = true): Promise<amqp.ConfirmChannel> {
```
francinelucca, line 46: should we add Jsdocs to the public functions here?

packages/api/src/main/messaging/MessagingClient.ts
```diff
@@ -0,0 +1,184 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+import amqp from 'amqplib'
+import { v4 as uuidv4 } from 'uuid'
+
+import { EventMessage, QueryMessage } from './config'
+import { MESSAGE_QUEUE_URL } from './constants'
+
+const RANDOM_QUEUE_NAME = ''
+const DEFAULT_ROUTING_KEY = ''
```
francinelucca, line 14: are these values meant to be empty?

packages/api/src/main/messaging/config.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+/**
+ * Each service gets a queue. Services bind their queue to one or more exchanges to receive
+ * messages.
+ */
+enum Queue {
+  Github = 'q_github_v1',
+  Logging = 'q_logging_v1',
+  Search = 'q_search_v1'
+}
```
francinelucca, line 15: are we using this anywhere?

packages/api/src/main/messaging/constants.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+const DEFAULT_EXCHANGE_TYPE = 'fanout'
+
+// Message queue URL checking
+if (getRunMode() === PROD && !process.env.CARBON_MESSAGE_QUEUE_URL) {
+  throw new Error('CARBON_MESSAGE_QUEUE_URL must be set in PROD run mode')
+}
+if (MESSAGE_QUEUE_URL === LOCAL_MESSAGE_QUEUE_URL) {
+  console.warn(`${chalk.bgYellowBright.black('WARN')} Using localhost message queue url`)
+}
```
francinelucca, line 32: is the constants file the correct place to do these checks? should we check the env vars in the constants file by standard from now on?

packages/api/src/main/messaging/constants.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export {
+  DEBUG,
+  DEFAULT_BIND_PATTERN,
+  DEFAULT_EXCHANGE_OPTIONS,
+  DEFAULT_EXCHANGE_TYPE,
+  DEFAULT_QUEUE_OPTIONS,
+  MESSAGE_QUEUE_URL
+}
```
francinelucca, line 41: seems like most of these aren't referenced anywhere, are they part of a separate PR?

packages/api/src/test/messaging/MessagingClient.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    mockedAmqp.connect.mockImplementation(() => {
+      throw new Error('connection failure test!')
+    })
+
+    const client = MessagingClient.getInstance()
+
+    await expect(client.connect(false)).rejects.toThrow()
+  })
+
+  it('quietly ignores replies for unknown correlation IDs', async () => {
```
francinelucca, line 144: wouldn't this test pass whether the correlation id was found or not? could we assert that replyCallbacks.delete wasn't called to ensure the function returned earlier?

package.json
jdharvey-ibm: I don't think so. I'll remove that.

packages/api/src/main/messaging/MessagingClient.ts
jdharvey-ibm: I did a docs run through all of these PRs. Definitely let me know if you have any comments on them or things you think I should add/change

packages/api/src/main/messaging/MessagingClient.ts
jdharvey-ibm: Yeah. I added some docs to these to clarify.

packages/api/src/main/messaging/config.ts
jdharvey-ibm: Not in this PR, but the logging one is used over [here](https://github.com/carbon-design-system/carbon-platform/blob/9ca0958b008ed2f72d573834971d6d87474495bb/services/logging/src/main/index.ts)

packages/api/src/main/messaging/constants.ts
jdharvey-ibm: Definitely not where we want to do these, haha. I'm basically just waiting on the stuff you have on your branch to make it in and we can do a pass through all of these and make them use the new way of orchestrating them.

jdharvey-ibm: Created issue #202

packages/api/src/main/messaging/constants.ts
jdharvey-ibm: They're used over in PR #193

packages/api/src/test/messaging/MessagingClient.test.ts
jdharvey-ibm: It would, yeah. The only purpose of this test is to ensure that a known non-existant value is handled gracefully. I'll add a comment to the test to clarify.

docs/packages-api.md
```diff
@@ -18,14 +18,17 @@ messages to be logged.
...
+- `PROD`
 
 Exporting the `CARBON_RUN_MODE` environment variable as either of these values will cause
-`getRunMode` to recognize it as the current run mode. Any other value will result in the mode being
-set to `DEV`.
+`getRunMode` to recognize it as the current run mode.
+
+If the environment variable is not set, the mode will default to `DEV`.
+
+Settings this envvar to any other value will result in an error being thrown.
```
francinelucca, line 34:
```suggestion
Setting this envvar to any other value will result in an error being thrown.
```

francinelucca (changes requested):

packages/api/src/main/messaging/MessagingClient.ts
francinelucca: LGTM!

.eslintrc.js
```diff
@@ -69,6 +69,13 @@ module.exports = {
           }
         ]
       }
+    },
+    {
+      files: ['*.ts'],
+      rules: {
+        'no-unused-vars': 'off',
+        '@typescript-eslint/no-unused-vars': ['error']
+      }
```
francinelucca, line 78: could the services/web-app/lib/file-cache.js sonarcloud error be caused by this change?

jdharvey-ibm: Nah, that was just me being dumb and editing a change on the github UI rather than in my IDE. I commented out the only use of a var, but didn't comment out the var definition itself.

This change is needed because the baseline eslint rule for unused vars doesn't properly handle typescript types and a couple of other weird edge cases.

francinelucca (dismissed):

francinelucca (dismissed):
LGTM!

francinelucca (dismissed):

docs/packages-api.md
```diff
@@ -18,14 +18,17 @@ messages to be logged.
...
 
-A package that exports a function called `getRunMode()`, which returns one of two constants,
+A module that exports a function called `getRunMode()`, which returns one of two constants,
 depending on the value of the `CARBON_RUN_MODE` environment variable.
 
 The possible values are:
 
 - `DEV`
-- `PRODUCTION`
+- `PROD`
```
mattrosno, line 27: Since we're now using this in the web-app, can you update from PRODUCTION to PROD in:

- [ ] `constants.js:7`
- [ ] `constants.js:9`

mattrosno (dismissed):
Thanks for the thorough documentation! I just read through the markdown.

docs/packages-api.md
jdharvey-ibm: Done!

francinelucca (approved):
LGTM

PR #192, opened 2022-01-14
feat(api): implement logging api

jharvey10:
This closes out the remainder of the v0 logging epic (along with #193)

Closes #84, Closes #90

PR #193, opened 2022-01-14
feat(logging): implement logging service

### LINKED ISSUE #91: Logging service: Connect to message broker (listen for events)

### LINKED ISSUE #92: Logging service: Implement message handler logic

### LINKED ISSUE #93: Logging service: Implement ingestion logic into LogDNA (IBM Log Analysis)

### LINKED ISSUE #94: Logging service: Unit tests

jharvey10:
This closes out the remainder of the v0 logging epic (along with #192)

Closes #91, Closes #92, Closes #93, Closes #94

Full implementation of the logging service to ingest logs and shuttle them to LogDNA


services/logging/src/test/logging.controller.test.ts
```diff
@@ -0,0 +1,85 @@
...
+      environment: 'test',
+      level: 'debug',
+      message: 'test message',
+      service: 'test service',
+      timestamp: Date.now()
+    }
+  })
+
+  it('logs a warning when no component specified', async () => {
+    delete (data as any).component
```
francinelucca, line 51: why do we need to cast to any?

services/logging/src/test/logging.controller.test.ts
```diff
@@ -0,0 +1,85 @@
...
+
+describe('message validation', () => {
+  let logDnaService: LogDnaService
+  let loggingController: LoggingController
+  let data: LogLoggedMessage
+
+  beforeEach(() => {
+    logDnaService = new LogDnaService()
+    loggingController = new LoggingController(logDnaService)
+    ;(loggingController as any).nestLogger = {
```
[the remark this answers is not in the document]
jdharvey-ibm: That would definitely be more clear as far as the intent of that part of the test, but unfortunately since `nestLogger` is marked as read-only in `PlatformController`, you can't set it when the object is asserted as a `PlatformController`

services/logging/src/test/logging.controller.test.ts
jdharvey-ibm: None of the fields on `data` are marked as optional (and they're all required in the map passed to `logLogged()`, so to get around this and fake an incoming request that is missing one of the fields, I first assert it as `any`, which gives unmitigated access to the object properties, delete one of the required fields and the pass it in.

As an alternative to this, I could instead have two interfaces. One called `LogLoggedMessage` and another called like `UnvalidatedLogLoggedMessage`, and only begin using `LogLoggedMessage` once validation has been performed successfully.

What are your thoughts on that as a general paradigm?

services/logging/src/main/decorators/Trace.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright IBM Corp. 2021, 2021
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import { PlatformController } from '../common/platform.controller'
+
+// TODO: incorporate messaging into this and make it more general purpose
+
+function Trace(): MethodDecorator {
```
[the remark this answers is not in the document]
francinelucca: I don't see an `@trace` usage anywhere, is nestjs using this implicitly?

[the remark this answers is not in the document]
jdharvey-ibm: No, nobody is using it at the moment. It's something I know we'll want to use in other services, so I wanted to keep the code around since it took a bit of research to figure out how to get it working properly, but at the moment it's not being used anywhere. Eventually we should move it to a common place along with the PlatformController, Validate, and PlatformMicroservice.

services/logging/src/test/logging.controller.test.ts
francinelucca: oh ok I understand now. I don't think having the two interfaces is necessary we can leave like this unless you really want to change it

francinelucca (dismissed):

mattrosno (dismissed):
Only read through the markdown changes and looked for web app side effects. Found none.

services/logging/src/test/logging.controller.test.ts
jdharvey-ibm: I think I am gonna do two actually haha. I think this would be a good case for using type assertions on methods like you and I chatted about the other day. So the validator function would become like:

```ts
function validate(data: UnvalidatedMessage): data is LogLoggedMessage {
  // do all the checks to ensure data is a valid LogLoggedMessage
  // and `throw` otherwise
}
```

I just need to wait until #192 is merged so I can have access to LogLoggedMessage on this branch.

francinelucca (dismissed):
lgtm

francinelucca (approved):

PR #199, opened 2022-01-20
feat: add env var to include ghe repos

mattrosno:
Adds an environment variable to conditionally load IBM internal content. This allows us to temporarily host on Vercel to have PR deploy previews.

Because `libraryAllowList` is just used in files ran on Next.js server, we can safely use `GITHUB_IBM_ACCESS` env var and not leak that value or any of the libraries in the client. The client-side JS bundle doesn't have the `libraryAllowList` object because it's not used in any client-side JavaScript.

#### Changelog

**New**

- `GITHUB_IBM_ACCESS` optional env var

**Changed**

- Functionality to use all or some of the libraries

**Removed**

- N/A

#### Testing / reviewing

1. Don't add the new env var, clean, run dev, and you should not see IBM Cloud PAL and the Watson Moments libraries
2. Add `GITHUB_IBM_ACCESS=true`, clean, run dev, and you should see all libraries like before


mattrosno:
Going with a different option after talking with @jdharvey-ibm and @francinelucca.

PR #200, opened 2022-01-21
chore: save prototyping data

mattrosno:
Commits prototyping data to this repository so we can host the prototype on Vercel without Vercel needing to make GitHub API requests to either GitHub public or enterprise.

This results in 498 JSON files (9.2MB) being committed to the repository.

#### Changelog

**New**

- `/data/libraries.js` specifies Carbon charts and components libraries to include
- `/config/constants.js` uses the platform run time, but defaults to `.carbon-oss`, so Vercel shouldn't need any env vars set

**Changed**

- By defaults uses the new `.carbon-oss` file system cache directory. To include the full data set, set `CARBON_RUN_MODE=PRODUCTION`

**Removed**

- N/A

#### Testing / reviewing

1. Run clean and dev. You should only see charts and components libraries.
2. Set `CARBON_RUN_MODE=PRODUCTION`, clean, dev, and you should see all libraries like before.
3. PRO MODE TEST turn off wifi / internet and do a build (it works, meaning it isn't reaching out to GitHub APIs anywhere)


services/web-app/config/constants.js
```diff
@@ -4,5 +4,5 @@
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
  */
-export const CACHE_PATH = '.carbon'
+export const CACHE_PATH = process.env.FILE_SYSTEM_CACHE || '.carbon-oss'
```
jdharvey-ibm: Since we don't expect these to change, I'd recommend:

```js
import { getRunMode, DEV, PROD } from '@carbon-platform/api/run-mode'

export const CACHE_PATH = getRunMode() === PROD ? '.carbon' : '.carbon-oss'
```

services/web-app/data/libraries.js
```diff
@@ -289,6 +289,37 @@ const libraryAllowList = {
...
+  'carbon-web-components'
+]
+
+/**
+ * Only use the `ossLibraries` data set if specified in the environment variable.
+ */
+const libraryAllowList = Object.keys(libraries)
+  .filter((key) =>
+    process.env.FILE_SYSTEM_CACHE !== '.carbon' ? ossLibraries.includes(key) : true
+  )
```
jdharvey-ibm: Similar comment here about using run-mode
```suggestion
  .filter((key) =>
    getRunMode() === DEV ? ossLibraries.includes(key) : true
  )
```

francinelucca (commented):
- [x] Ran without env var and only see select list of libs as expected
- [x] Ran with FILE_SYSTEM_CACHE=.carbon and can see all libs and components as expected
- [x] Build with FILE_SYSTEM_CACHE-.carbon and no wifi renders no libs and no components as expected
- [x] Build with FILE_SYSTEM_CACHE=.carbon-oss and no wifi renders subset of libs and components as expected

mattrosno:
Thanks for the thorough test and linting fixes @francinelucca and @jdharvey-ibm with the run mode suggestion.

New commit implements run mode. If you want to build with the full data set, use `CARBON_RUN_MODE=PRODUCTION` in your local env file.

jdharvey-ibm:
👍 Looks good to me. Side note: @francinelucca has some interesting stuff on the way with regards to environment variables and other config settings that should make tracking all of our various envvars a bit simpler.

jharvey10 (approved):

PR #205, opened 2022-01-24
feat(api): introduce enforce env vars package with getEnvVar function

### LINKED ISSUE #198: Enhance enforce-env-vars package
### Summary

Two additional functions:
`getEnvVar(name_of_var)`
and
`getEnvVar(name_of_var, fallback_value)
`
That will make the interface very similar to Java's `Properties` utilities.

That way, we can pull these from environment variables or even from another source like a config service or database or something in the future. Providing a fallback value would effectively do what the throwError boolean does in enforceEnvVars, whereby if there's a fallback, it won't throw an error unless you're in PROD mode. And if no fallback is specified, it will always throw an error if the var can't be found.

Self-managing these functions in this file will also allow us to cache values if we decide that's something we need.

### Justification

_No response_

### Success metrics

_No response_

### Required functionality

_No response_

### Specific timeline issues / requests

_No response_

### Available resources

_No response_

francinelucca:
Closes #198 

Decided to go against method overloading as research shows it's not a good practice, see https://dmitripavlutin.com/typescript-function-overloading/#4-when-to-use-function-overloading


packages/api/src/main/enforce-env-vars/index.ts
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright IBM Corp. 2022, 2021
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { getRunMode, PRODUCTION } from '../run-mode'
```
mattrosno, line 8: When @jdharvey-ibm's PR gets merged, this will need updated to `PROD`. Just an FYI.

mattrosno (approved):

jharvey10 (approved):

PR #213, opened 2022-01-26
fix(micromanage): add shell command guards

jharvey10:
- Also update some vars and terminology in deploy script

Addresses a SonarCloud security hotspot


francinelucca (approved):
LGTM!

PR #215, opened 2022-01-26
fix: prevent prototyping data from invalidating

mattrosno:
If not in PROD mode, meaning we're using the committed prototyping data, use a one year time to live in the cache so it doesn't become stale, dirtying our pull requests.

In doing so, all ~500 cache files are updated here when I ran development mode. Once this is merged into main and then into our PRs, our PRs will no longer have expired caches.


francinelucca (dismissed):
confirmed files are not being overwritten when running on DEV or PROD, LGTM!

jharvey10 (dismissed):

jharvey10 (approved):

PR #221, opened 2022-01-28
fix(api): minor auth updates

jharvey10:
- move dev store deps to be normal production deps
- use dynamic imports in store instead of require statements
- adjust manner in which paths are obtained for sqlite store
- clean  up test-generated .dev folder with sqlite store after running test suite


francinelucca (dismissed):
LGTM!

PR #231, opened 2022-02-01
Flucca/enhancement/local passport strategy

### LINKED ISSUE #212: Auth: Mock Local Authentication (DEV mode)
in PROD, you have to use https, but in DEV mode, you have the option of whether or not you want to run it in secure mode and still be able to fiddle with auth stuff.

The store stuff and all that should in theory be able to exist as it is. The difference would be that in DEV mode, the passportjs stuff could use a different provider besides the OIDC one. It could instead be one that pulls from local data or just works off of a hard coded value someplace:
PROD mode - always https. uses mongodb. db url obtained via envvar. uses ibm oidc provider url/secrets from envvars. works the same regardless of if you run PROD locally or PROD on code engine.
DEV mode - can use https. can also not use https. uses local sqlite db. can use local auth mechanism instead of oidc. can also instead use oidc if envvar is provided pointing to it.

francinelucca:
Closes #212


jharvey10 (approved):
We chatted a bunch about ways to scale this out and simplify some stuff which we will handle in separate branches.

PR #238, opened 2022-02-04
feat(web-app): new prototyping asset thumbnails

mattrosno:
From @jeanservaas to verify the new design kit to make thumbnails has the correct drawing area so nothing is clipped at grid or list views across all of our breakpoints.

In the future, when pulling these in from remote GitHub repos along with the metadata yaml files, we can optimize the SVGs at web app build time.


mattrosno:
@carbon-design-system/carbon-platform-devs merging this. Just new SVGs no source code changes.

PR #240, opened 2022-02-04
feat(web-app): introduce auth middleware, protect all routes

francinelucca:
Will merge once prod auth setup is finalized


services/web-app/package.json
```diff
@@ -16,7 +16,7 @@
     "lint": "next lint",
     "service:build": "npm run build",
     "start": "next start",
-    "start-secure": "sh ./scripts/create-local-certificates.sh && node server.js",
+    "start-secure": "sh ./scripts/create-local-certificates.sh && RUNNING_SECURELY=1 NODE_TLS_REJECT_UNAUTHORIZED=0 node server.js",
```
jdharvey-ibm, line 19: minor: consider renaming the npm script to `start:secure` to match our other services.

services/web-app/pages/_middleware.js
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { NextResponse } from 'next/server'
+
+import { isValidIbmEmail } from '@/utils/string'
+
+import { getRunMode, RunMode } from '../../../packages/api/runtime'
```
jdharvey-ibm, line 12:
```suggestion
import { getRunMode, RunMode } from '@carbon-platform/api/runtime'
```

services/web-app/pages/_middleware.js
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Creates request options for retrieveUser api call including headers and agent if necessary
+ *
+ * @param {NextRequest} req getServerSideProps request object
+ * @returns {RequestInit} request options
+ */
+function getRequestOptions(req) {
+  let cookies = ''
+  Object.entries(req.cookies).forEach(([key, val]) => {
+    cookies += `${key}=${val}; `
```
jdharvey-ibm, line 23: Rather than building this value up by hand, let's instead use [cookie](https://www.npmjs.com/package/cookie). Or extract only the cookie value we know we should have, which would be the session id.

services/web-app/pages/_middleware.js
```diff
@@ -0,0 +1,49 @@
...
+  let user
+  if (userResponse.ok) {
+    user = await userResponse.json()
+  }
+  if (!isValidIbmEmail(user?.email ?? '')) {
+    const url = req.nextUrl.clone()
+    url.pathname = '/404'
+    return NextResponse.rewrite(url)
+  }
+  return NextResponse.next()
```
jdharvey-ibm, line 48:
```suggestion
  // Guard - non-200 user api response
  if (!userResponse.ok) {
    return exitWith404(req)
  }
  
  const user = await userResponse.json()
  
  // Guard - no valid user
  if (!user?.email) {
    return exitWith404(req)
  }
  
  // Guard - not a valid IBMer
  if (!isValidIbmEmail(user.email)) {
    return exitWith404(req)
  }
  
  return NextResponse.next()
```

My general philosophy with permissions model and security stuff is to keep the logic as flat as possible with the final "success" statement at the very bottom of the function. (Ideally no more than +1 cognitive complexity for the entire method). Each check or "guard" above that final line should handle one possible error scenario that prevents the success statement from being reached.

jharvey10 (changes requested):

jharvey10 (dismissed):

PR #241, opened 2022-02-04
docs: add prop types

mattrosno:
Adds prop types to all pages and components to help document our code and also proactively catch type errors while developing.

#### Changelog

**New**

- Prop types to all page and components

**Changed**

- Spread props in the `RequireAuth` component please review @francinelucca 

**Removed**

- The `sunset` value in JSDoc typedefs, as that's no longer part of the schema

#### Testing / reviewing

Run the web app and watch the console and terminal for prop type warnings. The only warning should be from Carbon's `AspectRatio` component that is missing some values that it supports.


services/web-app/components/catalog-results/catalog-results.js
```diff
@@ -22,4 +23,8 @@ const CatalogResults = ({ assets = [] }) => {
   )
 }
 
+CatalogResults.propTypes = {
+  assets: PropTypes.array
```
mattrosno, line 27: Fixing this to be an array of assets.

mattrosno: Done.

services/web-app/components/auth/require-auth.js
```diff
@@ -4,12 +4,20 @@
...
  * LICENSE file in the root directory of this source tree.
  */
-const RequireAuth = (props) => {
-  if (props.isAuthorized) {
-    return { ...props.children }
+import PropTypes from 'prop-types'
+
+const RequireAuth = ({ children, fallback: Fallback, isAuthorized }) => {
+  if (isAuthorized) {
+    return { children }
```
francinelucca, line 11:
```suggestion
    return children
```

services/web-app/components/auth/require-auth.js
```diff
@@ -4,12 +4,20 @@
...
   } else {
-    return <props.fallback />
+    return <Fallback />
   }
 }
 
+RequireAuth.propTypes = {
+  children: PropTypes.node,
+  fallback: PropTypes.element,
```
francinelucca, line 19:
```suggestion
  fallback: PropTypes.func,
```

francinelucca (changes requested):

mattrosno:
@francinelucca thanks for the review. I didn't exactly know how to test that auth component, so I guessed that fallback prop would be of type "element" guessing it was a React component passed in.

Docs: https://www.npmjs.com/package/prop-types

```javascript
  // A React element (ie. <MyComponent />).
  optionalElement: PropTypes.element,
```

francinelucca:
> @francinelucca thanks for the review. I didn't exactly know how to test that auth component, so I guessed that fallback prop would be of type "element" guessing it was a React component passed in.
> 
> Docs: https://www.npmjs.com/package/prop-types
> 
> ```js
>   // A React element (ie. <MyComponent />).
>   optionalElement: PropTypes.element,
> ```

Yeah I thought so too but was giving me an error something like `Invalid prop "Fallback" of type "function" supplied to "RequireAuth", expected a single ReactElement.`
I believe because we're passing a component type `FourOhFour` instead of a component instance `<FourOhFour />`. Anyways, I tested it with Proptypes.func and the warning is gone, everything else LGTM!
image

francinelucca (approved):

mattrosno:
@francinelucca got it. Thanks for the thorough review!

PR #246, opened 2022-02-09
[V0.1 Release] - chore(api): temporarily introduce CARBON_USE_LOCAL_DB

francinelucca:
allows overriding remote db usage for auth api for v0.1 release

### DIFF
--- a/packages/api/src/main/auth/config/constants.ts
+++ b/packages/api/src/main/auth/config/constants.ts
@@ -18,16 +18,18 @@ const {
   // Client secret tied to App registration on SSO provisioner (get this from dev team)
   CARBON_IBM_ISV_CLIENT_SECRET,
   // Strategy name that passport should use when authenticating (get this from dev team)
-  PASSPORT_STRATEGY_NAME
+  PASSPORT_STRATEGY_NAME,
+  // URL of session storage DB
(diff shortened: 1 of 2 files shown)


packages/api/src/main/auth/config/constants.ts
```diff
@@ -18,19 +18,17 @@ const {
...
-  CARBON_IBM_ISV_CLIENT_SECRET: 'none',
-  CARBON_MONGO_DB_URL: '',
-  CARBON_MONGO_DB_NAME: ''
+  CARBON_IBM_ISV_CLIENT_SECRET: 'none'
 })
 
+// TODO: incorporate these consts into loadEnvVars when mongo db is set up
+const CARBON_MONGO_DB_URL = ''
+const CARBON_MONGO_DB_NAME = ''
+
```
jdharvey-ibm, line 31: Let's leave these vars in the load statement and also include the `CARBON_USE_LOCAL_DB` in there as well. In CE, we can just set the mongo ones to dummy values and the use local db one to `1`. The mongo ones won't get used and it'll force us to update the code once we have mongo and unset that use local db envvar.

jharvey10 (changes requested):

jharvey10 (approved):

PR #254, opened 2022-02-14
fix(web-app): override carbon overlay z-index

### LINKED ISSUE #249: Z index issue with the side nav overlay at medium breakpoint
### Browser

_No response_

### Description

We need to fix our `z-index`s to not conflict with the Carbon side nav overlay. Either reduce our z-index values and document those, or increase the z-index for Carbon's side nav overlay.

image

### Steps to reproduce

Medium breakpoint, catalog page, open side nav.

francinelucca:
Closes #249

### DIFF
--- a/services/web-app/layouts/layout/layout.module.scss
+++ b/services/web-app/layouts/layout/layout.module.scss
@@ -14,4 +14,7 @@
   // position for PageBreadcrumb
   position: relative;
   margin: $spacing-09 0 0;
+  :global(.cds--side-nav__overlay.cds--side-nav__overlay-active) {
+    z-index: 10;
+  }
 }
--- a/services/web-app/layouts/layout/layout.module.scss
+++ b/services/web-app/layouts/layout/layout.module.scss
@@ -14,6 +14,7 @@
   // position for PageBreadcrumb
   position: relative;
   margin: $spacing-09 0 0;
+  // have carbon overlay use higher z-index than we're using on our custom components
   :global(.cds--side-nav__overlay.cds--side-nav__overlay-active) {
     z-index: 10;
   }


services/web-app/layouts/layout/layout.module.scss
```diff
@@ -14,4 +14,7 @@
   // position for PageBreadcrumb
   position: relative;
   margin: $spacing-09 0 0;
+  :global(.cds--side-nav__overlay.cds--side-nav__overlay-active) {
+    z-index: 10;
```
mattrosno, line 18: How'd you land on 10? Maybe just a personal preference, but I like to add a comment when I have to use a z-index besides -1, 0, or 1.

mattrosno (dismissed):

services/web-app/layouts/layout/layout.module.scss
francinelucca: we're currently using 0,1 & 2 in our custom components in the web-app, I just needed a number that was higher than that, figured I'd add some wiggle room.
Happy to change it to lower if you think necessary!
image

mattrosno: @francinelucca I'd either go 3 and add a comment why, or maybe use a Carbon z-index. I don't know how/where these are used. https://github.com/carbon-design-system/carbon/blob/48deedf2fb7a343594ed844d6c852c9108a3eb53/packages/components/src/globals/scss/_layout.scss#L14

mattrosno: Oh, that's a private mixin. I don't know if Carbon has z-index's as part of the public API. I'd go 3.

mattrosno: Oh again, but the function `z` is public. https://github.com/carbon-design-system/carbon/blob/48deedf2fb7a343594ed844d6c852c9108a3eb53/packages/components/src/globals/scss/_layout.scss#L35

mattrosno (approved):

PR #260, opened 2022-02-16
fix(web-app): query refactor

### LINKED ISSUE #223: Malformed or non-existent filters cause web-app UI to hang
### Browser

Chrome

### Description

Changing the filters to malformed or unexpected values causes an infinite loop in the client UI of the assets catalog.

### Steps to reproduce

Example filter values that cause the hang:

```
?filter=%7B%22sponsor%22%3A%5B%22carbon%22%5D%2C%22status%22%3A%5B%22stable%22%5D%2C%22hello%22%3A%5B%22world%22%5D%7D
```
```
?filter={"hello":["world"]}
```
```
?filter={"hello":"world"}
```
```
?filter=a
```

The failure looks to be ultimately coming from the `catalog-filters.js` file, but it looks like when this component throws an error, it causes some other state change higher in the component hierarchy that ends up causing an infinite loop.

Related enhancement:

- [ ] #313

francinelucca:
validate json object before parsing, introduce pretty object strategy

Closes #223 

Addresses JSON parsing issues on query, validates filter before using and introduces new serializing query strategy for objects ( ?obj[key1]=value1&obj[key2]=value2... instead of JSON.stringify(object) )

#### Testing / reviewing

Pull changes and run application, test that applying and clearing filters in components page works as expected, test navigating to the app with filters described on #223 or any other unexpected filter value and assert app does not break


services/web-app/utils/string.js
```diff
@@ -32,3 +32,17 @@ export const isValidIbmEmail = (str) => {
   const acceptableEmailSuffixes = ['.ibm.com', '@ibm.com']
   return acceptableEmailSuffixes.some((suffix) => str.endsWith(suffix))
 }
+
+/**
+ * checks if a given string can be parsed to JSON
```
mattrosno, line 37:
```suggestion
 * Checks if a given string can be parsed to JSON
```

services/web-app/utils/string.js
```diff
@@ -32,3 +32,17 @@ export const isValidIbmEmail = (str) => {
...
+ * @param {string} str string to evaluate
+ * @returns {boolean} true if string is a valid JSON
+ */
+export const IsJsonString = (str) => {
```
mattrosno, line 41:
```suggestion
export const isJsonString = (str) => {
```

services/web-app/utils/use-query-state.js
```diff
@@ -5,37 +5,99 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import { set } from 'lodash'
 import { useRouter } from 'next/router'
 import { useCallback, useEffect, useMemo } from 'react'
 
+import { IsJsonString } from './string'
```
mattrosno, line 12:
```suggestion
import { isJsonString } from './string'
```

services/web-app/utils/use-query-state.js
```diff
@@ -5,37 +5,99 @@
...
   isoDateTime: {
-    parse: (v) => new Date(v),
-    serialize: (v) => v.toISOString()
+    parse: commonParser((v) => new Date(v)),
+    serialize: commonSerializer((v) => v.toISOString())
   },
   object: {
-    parse: (v) => JSON.parse(v),
-    serialize: (v) => JSON.stringify(v)
+    parse: commonParser((v) => (IsJsonString(v) ? JSON.parse(v) : {})),
```
mattrosno, line 95:
```suggestion
    parse: commonParser((v) => (isJsonString(v) ? JSON.parse(v) : {})),
```

services/web-app/components/catalog/catalog.js
```diff
@@ -79,7 +80,7 @@ function Catalog({ collection, data, type, filter: defaultFilter = {}, glob = {}
   })
 
   const [filter, setFilter] = useQueryState('filter', {
-    ...queryTypes.object,
+    ...queryTypes.prettyObject,
     defaultValue: defaultFilter
```
mattrosno, line 84: As a safe guard, we might want to clean the filter in the two places that we set it. Here and in `handleFilter()`.

```suggestion
    defaultValue: getCleanFilter(defaultFilter)
```

services/web-app/components/catalog/catalog.js
```diff
@@ -134,7 +135,7 @@ function Catalog({ collection, data, type, filter: defaultFilter = {}, glob = {}
     setFilteredAssets(
       assets
         .sort(assetSortComparator(sort))
-        .filter((asset) => assetIsInFilter(asset, filter))
+        .filter((asset) => assetIsInFilter(asset, getCleanFilter(filter)))
```
mattrosno, line 138: Can we remove this `getCleanFilter` and the 3 below it, and instead, run it in the `handleFilter()` callback because that's the only place that the filter is set? Have that return `setFilter(getCleanFilter(updatedFilter))`?

mattrosno (commented):
Very cool. I wish brackets and parenthesis weren't escaped in URLs, but this: 

`filter%5Bframework%5D=angular&filter%5Bframework%5D=react&filter%5Bsponsor%5D=carbon&filter%5Bstatus%5D=stable&view=grid&sort=a-z`

is better than the previous:

`filter=%7B"framework"%3A%5B"angular"%2C"react"%5D%2C"sponsor"%3A%5B"carbon"%5D%2C"status"%3A%5B"stable"%5D%7D&view=grid&sort=a-z`.

I guess it's pretty much the same length because of the duplicated "filter"s.

@francinelucca can you add comments to the `use-query-state.js` functions? I was getting a little lost reviewing that.

Also, I think I want to review my `initialFilter` strategy to initialize the filter with type and collection for tags. My logic there is starting to smell with too much prop drilling.

jharvey10 (commented):
My gut feeling on this change is that it seems like a lot of custom for a problem that might already be solved in nodejs or in other libraries.

Did you look into any existing libs? One I see that is built into node is the `URLSearchParams` class, which seems like it might be a good fit. Since it's a class, it could even be extended to include custom validation based on the valid filters we use.

francinelucca:
@jdharvey-ibm we're using URLSearchParams in our implementation. We're trying to solve for having to read an object from the query, didn't find a whole lot of options online so Matt & I settled for the brackets rather than JSON.stringify. I can definitely revert back to the old way if desired or happy to jump on a call to discuss other possible strategies

jdharvey-ibm:
Ah, I see the usage of URLSearchParams you mentioned. I think the changes you've made are headed in the right direction because they do break up the filters into separate parameters, but I think you can go even further and rely almost entirely on the URLSearchParams class for its handling of objects, arrays, iteration, value type coercion, modification, sorting, and serialization.

The only logic on top of that base functionality would then just need to be ensuring that the filter params are valid catalog filters, which could either be done through a utility function or an extension of the URLSearchParams class to include a 'validate' method that can be called to perform all of that validation.

We can definitely chat more on the side about this though if you'd like.

francinelucca:
@mattrosno per our conv with @jdharvey-ibm yesterday we agreed not to make changes to the filter logic until we make the refactor, which we pushed out to next sprint so I took #223 out of the scope of v0.1 and will work on this next sprint

mattrosno (commented):
Functionally, this is working great. Any way to abstract some repetition into functions and add some more comments? I want to make a pass and add comments of my own to these files - the parts that you didn't touch.

I'll do more testing with edge cases later.

services/web-app/components/catalog/catalog.js
```diff
@@ -52,36 +54,141 @@ const assetIsInFilter = (asset, filter) => {
...
+      defaultValue: defaultFilter.framework
+    },
+    (value) => {
+      // assert framework value is an array and has valid values
+      return (
+        !!value &&
+        value.constructor === Array &&
+        !!possibleFilterValues?.framework?.values &&
+        !value.some((val) => !Object.keys(possibleFilterValues.framework.values).includes(val))
+      )
```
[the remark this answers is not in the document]
francinelucca: Refactored into `filterPropertyHasValidValue `

services/web-app/utils/use-query-state.js
```diff
@@ -6,47 +6,13 @@
...
-    parse: (v) => new Date(v),
-    serialize: (v) => v.toISOString()
-  },
-  object: {
-    parse: (v) => JSON.parse(v),
-    serialize: (v) => JSON.stringify(v)
-  }
-}
-
 export const useQueryState = (
```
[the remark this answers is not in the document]
francinelucca: Done!

services/web-app/components/catalog/catalog.js
```diff
@@ -153,8 +260,36 @@ function Catalog({ collection, data, type, filter: defaultFilter = {}, glob = {}
...
+  const updateQueryValues = (newFilter) => {
+    const queryUpdates = {}
+    if (!isEqual(newFilter.sponsor, filter.sponsor)) queryUpdates.sponsor = newFilter.sponsor
+    if (!isEqual(newFilter.platform, filter.platform)) queryUpdates.platform = newFilter.platform
+    if (!isEqual(newFilter.status, filter.status)) queryUpdates.status = newFilter.status
+    if (!isEqual(newFilter.tags, filter.tags)) queryUpdates.tags = newFilter.tags
+
+    if (!isEqual(newFilter.framework, filter.framework)) {
+      queryUpdates.framework = newFilter.framework
+    }
```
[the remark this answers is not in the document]
francinelucca: Went around this one for a while, don't think it's worth it to try to put this into a for loop or something of the sorts because the work that'd need to be done to figure out what `setSponsor()`, `setFramework()`, `setPlatform()` to call would result in way more lines (see new changes)

mattrosno (changes requested):
So cool to prevent those unnecessary full-page re-renders as the URL query string updates. Looks great! Here's a PR that we can review and merge into this PR and on my end, this looks good.

https://github.com/carbon-design-system/carbon-platform/pull/334

francinelucca:
@mattrosno merged, Thanks!🙏

mattrosno (dismissed):

mattrosno:
@francinelucca the code smell error: https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&types=CODE_SMELL&pullRequest=260&id=carbon-design-system_carbon-platform&open=AX9meF3lDu2tD1MhHatF

mattrosno (dismissed):
Ran locally and applied filters and search and it worked as expected 👍

services/web-app/utils/use-query-state.js
```diff
@@ -57,15 +28,35 @@ export const useQueryState = (
       return null
```
jdharvey-ibm, line 28: This comment on line 53 (old) 24 (new) probably needs an update now

services/web-app/utils/use-query-state.js
```diff
@@ -6,47 +6,18 @@
...
-  object: {
-    parse: (v) => JSON.parse(v),
-    serialize: (v) => JSON.stringify(v)
-  }
+import queryString from 'query-string'
+import { useCallback, useEffect, useState } from 'react'
+
+const queryStringConfig = {
+  arrayFormat: 'bracket-separator',
+  arrayFormatSeparator: '|'
```
jdharvey-ibm, line 14: Could you add a link in this to the query string docs that show how this array format works?

services/web-app/utils/use-query-state.js
```diff
@@ -57,15 +28,35 @@ export const useQueryState = (
...
+
+    const storageValue = saveToStorage
+      ? localStorage.getItem(`${router.pathname}:${key}`)
+      : undefined
+
+    let val = query[key]
 
-    return queryValue !== null ? parse(queryValue) : storageValue
-  }, [key, parse, router.pathname, saveToStorage])
+    if (val !== undefined && validateValue && !validateValue(val)) {
```
jdharvey-ibm, line 43: super extra safety here would be to do something like:
```suggestion
    if (val !== undefined && typeof validateValue === 'function' && !validateValue(val)) {
```

services/web-app/utils/use-query-state.js
```diff
@@ -57,15 +28,35 @@ export const useQueryState = (
...
+      // Change query state without rerendering page
+      history.replaceState(null, null, `?${queryString.stringify(query, queryStringConfig)}`)
+    }
+
+    if (val === undefined) {
+      val = storageValue
+    }
+
+    return val
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
jdharvey-ibm, line 56: Which deps does it want you to include that aren't currently in here?

services/web-app/utils/use-query-state.js
```diff
@@ -57,15 +28,35 @@ export const useQueryState = (
...
-    return queryValue !== null ? parse(queryValue) : storageValue
-  }, [key, parse, router.pathname, saveToStorage])
+    if (val !== undefined && validateValue && !validateValue(val)) {
+      val = undefined
+      delete query[key]
 
-  // Update the "state" when the router.query key changes
-  const value = useMemo(getValue, [getValue, router.query[key]])
+      // Change query state without rerendering page
+      history.replaceState(null, null, `?${queryString.stringify(query, queryStringConfig)}`)
```
jdharvey-ibm, line 48: It looks like the goal of this logic is to remove invalid values from the query string. Is that desirable? Typically I'd expect to see the input data (query string) remain unaltered and have the application logic validate/sanitize it and then provide the valid values to the rest of the application.

services/web-app/utils/use-query-state.js
```diff
@@ -57,15 +28,35 @@ export const useQueryState = (
...
+    }
+
+    if (val === undefined) {
+      val = storageValue
+    }
```
jdharvey-ibm, line 53: It looks like this is reverting to the value form local storage if the supplied value was either absent or failed validation. What happens in the case where you're _not_ saving to local storage and the value becomes invalid (after having previously become valid)? Would the application see "undefined" now for the value?

I think this hook would benefit from top-level documentation describing the exact contract it provides, including these types of edge cases.

services/web-app/utils/use-query-state.js
```diff
@@ -75,19 +66,21 @@ export const useQueryState = (
...
-      router.replace(`?${query.toString()}`, undefined, { shallow: true, scroll: false })
+      // Change query state without rerendering page
+      history.replaceState(null, null, `?${queryString.stringify(query, queryStringConfig)}`)
+
+      setValue(getValue())
     },
-    // The Next.js router updates `router.replace`, which should not trigger re-renders
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [getValue, key, serialize]
+    [getValue, key]
```
jdharvey-ibm, line 83: same question here about which deps are missing

services/web-app/utils/use-query-state.js
```diff
@@ -75,19 +66,21 @@ export const useQueryState = (
...
-      if (newValue === null || newValue === undefined) {
-        query.delete(key)
-      } else {
-        query.set(key, serialize(newValue))
-      }
+      query[key] = newValue
 
-      router.replace(`?${query.toString()}`, undefined, { shallow: true, scroll: false })
+      // Change query state without rerendering page
+      history.replaceState(null, null, `?${queryString.stringify(query, queryStringConfig)}`)
```
jdharvey-ibm, line 78: Does this mean that no matter how many times you update the query string, you'll only ever have one history entry? If so, is that desirable? A lot of catalog-type websites I use don't behave that way; they update the URL each time the filter is applied (sometimes with debouncing or explicit "apply" buttons).

services/web-app/utils/use-query-state.js
```diff
@@ -57,15 +28,35 @@ export const useQueryState = (
       return null
     }
 
-    const query = new URLSearchParams(window.location.search)
-    const queryValue = query.get(key)
-    const storageValue = saveToStorage ? localStorage.getItem(`${router.pathname}:${key}`) : null
+    const query = queryString.parseUrl(window.location.search, {
```
jdharvey-ibm, line 31: If the query string changes programmatically from outside of this hook, is the expected behavior that the components using the `userQueryState` hook will be unaware of those changes?

services/web-app/components/catalog/catalog.js
```diff
@@ -51,37 +66,155 @@ const assetIsInFilter = (asset, filter) => {
...
-    saveToStorage: true
-  })
+  const [sort, setSort] = useQueryState(
+    'sort',
+    {
+      defaultValue: 'a-z',
+      saveToStorage: true
+    },
+    (value) => {
+      return !!value && typeof value === 'string' && ['a-z', 'status'].includes(value)
```
jdharvey-ibm, line 140: `value` will be coerced to a boolean here automatically, so no need to include the `!!`.

services/web-app/components/catalog/catalog.js
```diff
@@ -51,37 +66,155 @@ const assetIsInFilter = (asset, filter) => {
...
   const [search, setSearch] = useState(query)
 
-  const [sort, setSort] = useQueryState('sort', {
-    defaultValue: 'a-z',
-    saveToStorage: true
-  })
+  const [sort, setSort] = useQueryState(
+    'sort',
+    {
+      defaultValue: 'a-z',
```
jdharvey-ibm, line 136: You could potentially define the valid sort values in a constant and then use the 0-index as the default, so like:
```js
const VALID_SORT_VALUES = ['a-z', 'status]
// ...
defaultValue: VALID_SORT_VALUES[0]
```

You could then also use that array below when checking if the `value` is included in that list.

jdharvey-ibm: Side question: are those possible sort values already defined in a constant someplace? They're needed in at least one other place to display them on the UI.

(Similar questions and comments for all of the various filter values)

jharvey10 (changes requested):

services/web-app/utils/use-query-state.js
francinelucca: if the user changs it and presses enter it will 'cause a re-render, which will inform all comps listening for updates from use-query-state to be notified of that. if another component changes the queryState right now without going through the hook then yes, the components would be unaware but it's not an expected use-case for components to be manipulating the query state directly. 
Thoughts on this?

services/web-app/utils/use-query-state.js
francinelucca: the update function only runs on a per-key basis so only one query string key should be updated at a time

services/web-app/utils/use-query-state.js
francinelucca: ok! I initially went with the "if this value is not valid for a key, remove it from the query" approach, but I can definitely leave the query string unaltered and ignore the invalid value instead

services/web-app/utils/use-query-state.js
jdharvey-ibm: That makes sense, but I was wondering (user-experience-wise) if a particular key is updated 3 or 4 times in a row, what would the user's browser history look like? Would they see a history entry for each filter update? Or would they see only a single history entry? I'd argue that they should see one history entry per key update so they can use their back button to traverse how they'd changed their filters over time.

services/web-app/utils/use-query-state.js
jdharvey-ibm: Okay, got it. You're right that that's not an expected use case and is therefore pretty low priority. It could potentially be resolved by registering an event that listens for changes to the URL/query string and updating the hook accordingly, but no need to include that.

services/web-app/utils/use-query-state.js
mattrosno: @jdharvey-ibm as a general practice, should we add comments next to any `// eslint-disable-next-line react-hooks/exhaustive-deps` to explain why we're disabling `react-hooks/exhaustive-deps`?

jdharvey-ibm: Yeah! That'd be awesome!

francinelucca: Agreed! will do

services/web-app/utils/use-query-state.js
francinelucca: correct, the value would be undefined, that makes sense to me, should we code for a different behavior? 
I'll add some top level comments regarding the hook's functionality and how it behaves 👍

services/web-app/utils/use-query-state.js
francinelucca: `parseNumbers`, `parseBooleans`. Should I add those to the deps array or continue to ignore them? I get a little confused about adding deps I don't really "care" about to the deps array

services/web-app/utils/use-query-state.js
francinelucca: Removed, unnecessary

services/web-app/utils/use-query-state.js
francinelucca: Didn't need any extra dependencies for this case, removed comment

services/web-app/utils/use-query-state.js
francinelucca: Done!

services/web-app/utils/use-query-state.js
francinelucca: Refactored to no longer delete invalid values and ignore them instead

services/web-app/utils/use-query-state.js
francinelucca: did a test with 
`
   useEffect(() => {
     setQuery('q1')
     setQuery('q2')
     setQuery('q3')
     setQuery('q4')
   }, [sort])
`
and tried checking `history.back()` and confirmed can navigate back to all the changed filters (`?q=3`, `?q=2`, `?q=1`)

services/web-app/components/catalog/catalog.js
francinelucca: Done!

services/web-app/components/catalog/catalog.js
francinelucca: Moved sort and view constants into `services/web-app/data/sort.js` and `services/web-app/data/view.js` and made the necessary adjustments were those were hard coded in

services/web-app/utils/use-query-state.js
francinelucca: Added JSDocs!

alisonjoseph (dismissed):
Ran locally and filters appear to be working correctly 🥳

services/web-app/data/sort.js
```diff
@@ -0,0 +1,18 @@
...
+export const sortItems = [
+  {
+    id: ALPHABETICAL_ORDER,
+    text: 'A–Z'
+  },
+  {
+    id: ORDER_BY_STATUS,
+    text: 'Status'
+  }
+]
```
jdharvey-ibm, line 18: It feels a little weird putting this constants file in a data folder in the project. Is there any better home for constants-type stuff in the web-app? (same comment for the view.js file below)

services/web-app/utils/use-query-state.js
jdharvey-ibm: Yeah, definitely add those. In general, it's best to be exhaustive with those deps since it helps react make better determinations about when to re-execute hooks. Effect hooks aren't so much about defining the deps you want to cause rerenders; they're instead about tracking dependencies and allowing react to take over from there. In the docs I think they say someplace that the effect hook deps list will potentially at some point go away and be generated programmatically, so it's best for us to be in alignment with their eslint rules in all possible places.

jharvey10 (changes requested):

services/web-app/utils/use-query-state.js
```diff
@@ -9,20 +9,43 @@ import { useRouter } from 'next/router'
...
+ * 4 - Supplying option `saveToStorage` = `true` will cause the value
+ * to be stored within browser's localStorage. This value will be used instead of
+ * the query string value if it happens to be invalid or undefined
+ * 5 - Calling the update function with a `null`, or `undefined` value
+ * will cause the localStorage entry to be removed (if any)
+ * @param {string} key Key to use in the query string
+ * @param {{defaultValue: string, saveToStorage: boolean,
+ * parseNumbers: boolean, parseBooleans: boolean}} options Extra config options
+ *
+ * @returns {{value, update}} Current value and update function
```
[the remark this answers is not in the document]
francinelucca: can we chat about this? I don't think I'm following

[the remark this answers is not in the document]
francinelucca: - if (validationFunction returns false) <- read from localStorage
- update() -> whatever is in the query, always save to localStorage
resetOnLoad: reset to default when read is not triggered from an update()

services/web-app/components/catalog/catalog.js
```diff
@@ -130,31 +271,33 @@ function Catalog({ collection, data, type, filter: defaultFilter = {}, glob = {}
...
-          if (search) {
-            return (
-              (name && name.toLowerCase().includes(search.toLowerCase())) ||
-              (description && description.toLowerCase().includes(search.toLowerCase()))
-            )
-          }
+    setFilteredAssets(getFilteredAssets(assets, filter, sort, search))
+    // Do not want to run this everytime the filter changes
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [assets, JSON.stringify(filter), sort, search])
```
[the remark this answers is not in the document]
francinelucca: using `fillter` in array deps instead now

services/web-app/utils/use-query-state.js
francinelucca: Done!

services/web-app/utils/use-query-state.js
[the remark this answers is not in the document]
francinelucca: Redefined hook contract and made discussed changes: now always saving to localStorage and resetting to default on new page navigation (can bypass this behavior by sending resetOnLoad = false (defaults to true).

services/web-app/components/catalog/catalog.js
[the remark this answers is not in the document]
mattrosno: The `JSON.stringify` was probably from me, before this PR to check for object equality.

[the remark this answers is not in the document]
mattrosno: `filter` object is probably getting reinstantiated somewhere, meaning it's a new object, meaning even if the values didn't change, the effect would still run. This was a hack (I admit) to prevent that.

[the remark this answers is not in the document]
mattrosno: @jdharvey-ibm @francinelucca We might want to use something like https://github.com/streamich/react-use/blob/master/docs/useDeepCompareEffect.md for deep object comparisons in useEffect dependency arrays.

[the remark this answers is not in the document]
francinelucca: @mattrosno ended up going for a `usePrevious` implementation to check if the filter actually changed before running the useEffect code , you can view that specific change here: https://github.com/carbon-design-system/carbon-platform/commit/b2b2e57af024850957cb62d3a9d5b6ccc6422808

mattrosno (dismissed):

services/web-app/utils/use-query-state.js
```diff
@@ -21,58 +21,63 @@ const queryStringConfig = {
...
 
-    if (val === undefined) {
-      val = storageValue
-    }
+      if (typeof validateValue === 'function' && !validateValue(val)) {
+        const storageValue = localStorage.getItem(`${router.pathname}:${key}`)
+        if (parseBooleans) {
+          val = storageValue === 'true'
+        } else if (parseNumbers) {
+          val = Number(storageValue)
```
jdharvey-ibm, line 71: I'm not sure these are working as intended. On line 69, if the value coming out of local storage is `"a-z"` (or any string other than "`true`"), val will always be returned as `false` when `parseBooleans` is `true` since the string doesn't pass this equality test and isn't meant to be interpreted as a boolean. The same thing would happen on line 71 interpreting things as numbers.

Food for thought:
For a param coming out of localstorage that you want to convert to a string or a boolean:
1. get the value out of local storage
2. invoke `querystring.stringify(...)` to convert it to a url search string
3. invoke `querystring.parse(..., {parseBooleans, parseNumbers})` against that string
4. get the value back out of the returned result and it will now be cast to the appropriate type.

Something like:

```js
const storageVal = localStorage.getItem(key)
const storageSearchQuery = queryString.stringify({key: storageVal})
const parsedStorage = queryString.parse(storageSearchQuery, {parseBooleans, parseNumbers})
const parsedActual = queryString.parse(window.location.search, {parseBooleans, parseNumbers})
if (parsedActual[key] passes validation) {return parsedActual[key]}
else {return parsedStorage[key]}
```
Note that in this example, the local storage must only contain a value that has already been shown to pass the provided validation function (this is already being done in the code you have below)

jharvey10 (changes requested):

services/web-app/utils/use-query-state.js
```diff
@@ -21,58 +21,63 @@ const queryStringConfig = {
...
- * 3 - The `validateValue` function receives the current query string value
+ * 3 - All queryState key/values will be saved to localStorage
+ * 4 - The `validateValue` function receives the current query string value
  * and should return true if the value is valid or false otherwise. If the value is invalid,
- * the hook will return a value of `undefined`
- * 3- Supplying options `parseNumbers` or `parseBoolean` = `true` will cause the type of value
+ * the hook will return the last value saved to localStorage (null if not set)
+ * 5 - By default, queryStates are always saved to localStorage;
+ * supplying option `resetOnLoad` = `false` will cause the value to be removed from localStorage
+ * everytime a new page navigation to the path occurs
```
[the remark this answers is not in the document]
francinelucca: so we want the default behavior to be everything always stays in localStorage unless the consumer explicitly says not to? it seems to me that the most common use case will be you want the resetOnLoad

services/web-app/utils/use-query-state.js
```diff
@@ -21,58 +21,63 @@ const queryStringConfig = {
...
+      const query = queryString.parseUrl(window.location.search, {
+        ...queryStringConfig,
+        parseNumbers,
+        parseBooleans
+      }).query
 
-    let val = query[key]
+      if (!isInternalCall && resetOnLoad) {
+        localStorage.setItem(`${router.pathname}:${key}`, defaultValue)
+      }
```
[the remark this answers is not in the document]
francinelucca: Tried this but it's causing a lot of race conditions due to useEffect() running *after* the first render. See: https://reactjs.org/docs/hooks-effect.html#:~:text=Does%20useEffect%20run%20after%20every,render%20and%20after%20every%20update.

Left as is for now, open to any ideas

services/web-app/utils/use-previous.js
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright IBM Corp. 2022, 2022
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const { useRef, useEffect } = require('react')
+
+// Hook, see: https://usehooks.com/usePrevious/
```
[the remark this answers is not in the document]
francinelucca: Done!

services/web-app/utils/use-previous.js
```diff
@@ -0,0 +1,23 @@
...
+function usePrevious(value) {
+  // The ref object is a generic container whose current property is mutable ...
+  // ... and can hold any value, similar to an instance property on a class
+  const ref = useRef()
+  // Store current value in ref
+  useEffect(() => {
+    ref.current = value
+  }, [value]) // Only re-run if value changes
+  // Return previous value (happens before update in useEffect above)
+  return ref.current
```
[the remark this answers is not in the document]
francinelucca: Done!

services/web-app/utils/use-query-state.js
francinelucca: Done! did a variation of this, can see in 4a9221ec887a36f426571900246a316ec8c78963

mattrosno:
@francinelucca does https://github.com/carbon-design-system/carbon-platform/pull/420 need merged before this?

francinelucca:
> @francinelucca does #420 need merged before this?

either or, doesn't really matter

mattrosno:
@francinelucca can you resolve these conflicts then we get this PR reviewed and hopefully merged? Since this was carryover from last sprint, would be great to get in this sprint.

jharvey10 (approved):

andreancardona (approved):
🎉 looks good!

PR #267, opened 2022-02-18
feat: remove unused web app nav items

### LINKED ISSUE #252: Private beta fit and finish
- [ ] Remove standards page and nav
- [ ] Remove search and site switcher from global nav (icons that don't do anything yet)
- [ ] Remove elements catalog?
- [ ] If production environment, catalog items should only take you offsite (external URL) or be read-only, because library and asset detail pages are outside the scope of the private beta (we can keep those pages around, just not link to them from the interface)

mattrosno:
Closes #252 

#### Changelog

**New**

- `/` redirects to `/assets` because there is no home page yet and we might as well leave everything under the `/assets` route

**Changed**

- Side nav now has headings for groups instead of menus

**Removed**

- Standards and unused icons in global nav
- Elements from side nav and schema

#### Testing / reviewing

You still have to click the global nav user icon before seeing any page... the web app now looks like this:

image


francinelucca (approved):
LGTM! opening the app redirects me to /assets, still able to login and can't see elements option or removed icons

PR #269, opened 2022-02-18
chore: add resource schema package with updated schema files

### LINKED ISSUE #66: Add resource schema validation tooling
As mentioned by @jdharvey-ibm [here](https://github.com/carbon-design-system/carbon-platform/pull/65#issuecomment-958137019), we can create JSON schema files that can get referenced in metadata files to validate types, required fields, have auto-completion, etc.

- [x] JSON schema validation files for our main branch schema (before changes via @mattrosno)
- [x] Add repo package for schema JSON files that get versioned (so schema file names don't need versioned) (and so we could use unpkg or something else for prettier URLs to apply the schema validation to all indexing)

francinelucca:
Closes #66 

Adds resources-schema package (published on npmjs as @carbon-platform/schemas) with updated asset & library schema (single file) as per current docs.

#### Testing / reviewing

Schema file can now be consumed as: 

```yml
# yaml-language-server: $schema=https://unpkg.com/@carbon-platform/schemas/carbon-resources.schema.json
```

See example:

carbon.yml

```yml
# yaml-language-server: $schema=./schemas/carbon-resources.schema.json
library:
  id: lib-id
  name: lib-name
  demoLinks:
    - type: codesandbox
      name: somethng
      action: link
      url: "https://test.com"
  externalDocsUrl: "!"
  private: true
  packageJsonPath: "/package.json"
  description: "something"
assets:
  asset-id:
    name: asset-name
    description: asset-description
    status: stable
    type: component
    tags:
      - content-block
      - content-element
    framework: react
    platform: web
    thumbnailPath: /something
    externalDocsUrl: sample-url
    inherits:
      asset: inherited-asset-name
      properties:
        - name
      primary: false
    private: false
    demoLinks:
      - type: codesandbox
        name: something here
        action: link
        url: https://something
```


packages/resource-schemas/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@carbon-platform/resource-schemas",
```
mattrosno, line 2: Thoughts on shorting this package name to `@carbon-platform/schemas`? So in a future state once we update the schema file name, we should shorten the comment to this?

```yml
# yaml-language-server: $schema=https://unpkg.com/@carbon-platform/schemas/library.schema.json
```

packages/resource-schemas/carbon-asset.schema.json
```diff
@@ -0,0 +1,202 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://carbondesignsystem.com/schemas/carbon-asset.schema.json",
```
mattrosno, line 3: How does this `$id` get used? Since we aren't hosting these on cds.com, should use use the unpkg URLs?

packages/resource-schemas/carbon-asset.schema.json
```diff
@@ -0,0 +1,202 @@
...
+    "title": "Carbon Asset",
+    "description": "A Carbon Asset",
```
mattrosno, line 5: From our markdown file:

> Assets are reusable units of work that are used in products and digital experiences. Every asset belongs to a library.

packages/resource-schemas/carbon-library.schema.json
```diff
@@ -0,0 +1,82 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://carbondesignsystem.com/schemas/carbon-library.schema.json",
```
mattrosno, line 3: Same here.

packages/resource-schemas/carbon-library.schema.json
```diff
@@ -0,0 +1,82 @@
...
+    "title": "Carbon Library",
+    "description": "A Carbon Library",
```
mattrosno, line 5: > Libraries are the means to contribute, install, and use one or many assets in products and digital experiences.

mattrosno (commented):
Some small comments. Main thoughts are on shortening package and file names.

packages/resource-schemas/package.json
francinelucca: Done!

jharvey10 (changes requested):
Overall looks pretty good! One thing you need to do is to add the schemas package to the top-level package.json workspaces section and then run `npm install` to update the lock file.

mattrosno (dismissed):

mattrosno (dismissed):

.commitlintrc.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
         'micromanage',
         'release',
         'search',
+        'schemas',
```
jdharvey-ibm, line 20: bump this up by one to alphabetize :)

jharvey10 (changes requested):

jharvey10 (approved):

PR #276, opened 2022-02-23
(web-app): login page

### LINKED ISSUE #268: Log in interface
We are currently leveraging Next.js middleware to enforce IBMid authentication prior to viewing **any** page in the private beta. This is a short-term solution so we can show IBM internal content in the private beta. A later monthly release will have proper authentication where open-source content is visible prior to logging in.

Right now it looks like this when the page loads, then you click the user icon top right to log in, then you're in.

image

# Option 1

We add a log in button with copy to that 404 page. A simple "page not found, log in to view all pages." This is by far the simplest approach. We could even modify that 404 page to not show the log in prompt if you've since logged in.

# Option 2

We build a full login page (similar to https://badges.carbondesignsystem.com) where we take over the full page and no side nav and we prompt people to log in. I don't think it's worth that effort.

# Option 3

Some log in modal that pops up. There will still be an empty 404 page behind it, so this doesn't seem ideal.

# Recommendation

Option 1 to help us hit this first private beta release as it's the quickest to do and with minimal effort we can make that 404 page look good with the log in prompt.

francinelucca:
Closes #268 

**Testing/Reviewing**

Pull changes and run, make sure app looks as expected at different breakpoints as defined in #268 and that login prompt and button are not displayed when user is logged in.

### DIFF
--- a/services/web-app/components/PageNotFound/PageNotFound.js
+++ b/services/web-app/components/PageNotFound/PageNotFound.js
@@ -6,48 +6,44 @@
  */
 import { Button, Column, Grid } from '@carbon/react'
 import { ArrowRight } from '@carbon/react/icons'
+import { useEffect, useState } from 'react'
 
-// import { useAuth } from '@/contexts/auth'
-// import { isValidIbmEmail } from '@/utils/string'
-// import RequireAuth from '../auth/require-auth'
+import { useAuth } from '@/contexts/auth'
+import { isValidIbmEmail } from '@/utils/string'
+
 import styles from './PageNotFound.module.scss'
 
 export const PageNotFound = () => {
-  //  const { isAuthenticated, loading, user } = useAuth()
+  const [isAuthorized, setisAuthorized] = useState(false)
+  const { isAuthenticated, loading, user } = useAuth()
 
-  // const loginGrid =
+  useEffect(() => {
+    if (!loading && isAuthenticated && isValidIbmEmail(user?.email ?? '')) {
+      setisAuthorized(true)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [loading, isAuthenticated, user])
 
   return (
-    <>
-      {/* <RequireAuth
-      fallback={FourOhFour}
-      isAuthorized={isAuthenticated && isValidIbmEmail(user?.email ?? '')}
-    >
-      <>
-        Welcome to the Protected Page!
-        <div>User: {JSON.stringify(user ?? {})}</div>
-        <div>isAuthenticated: {isAuthenticated?.toString()}</div>
-        <div>isLoading: {loading?.toString()}</div>
-      </>
-    </RequireAuth> */}
-      <Grid className={styles.grid}>
-        <Column className={styles.column} sm={4} md={6} lg={6} xlg={6} max={6}>
-          <h1 className={styles.title}>Page not found.</h1>
+    <Grid className={styles.grid}>
+      <Column className={styles.column} sm={4} md={6} lg={6} xlg={6} max={6}>
+        <h1 className={styles.title}>Page not found.</h1>
+        {!isAuthorized && (
           <Grid>
             <Column sm={4} md={8} lg={12} xlg={12} max={12}>
               <h1 className={styles.title}>Log in to view all pages.</h1>
+              <Grid columns={3}>
+                <Column sm={3} md={2} lg={2} xlg={2} max={2}>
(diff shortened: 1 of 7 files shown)


services/web-app/components/PageNotFound/PageNotFound.module.scss
```diff
@@ -0,0 +1,24 @@
...
+  @include breakpoint-down(md) {
+    padding-right: calc($spacing-05 + var(--cds-grid-gutter) / 2);
+    padding-left: calc($spacing-05 + var(--cds-grid-gutter) / 2);
+  }
+}
+
+.loginButton {
+  width: 100%;
+  max-width: none;
+  margin-top: $spacing-07;
```
mattrosno, line 23: I think you need to add a `padding-right: $spacing-05;` to fix arrow positioning.

services/web-app/components/PageNotFound/PageNotFound.js
```diff
@@ -0,0 +1,47 @@
+/*
```
mattrosno, line 1: We follow kebab case for component file names. I'm pretty sure we made this decision based on Next.js conventions. Can you update?

services/web-app/components/PageNotFound/PageNotFound.module.scss
```diff
@@ -0,0 +1,24 @@
...
+
+.grid {
+  margin-top: $spacing-10;
+  @include breakpoint-down(md) {
+    padding-right: calc($spacing-05 + var(--cds-grid-gutter) / 2);
+    padding-left: calc($spacing-05 + var(--cds-grid-gutter) / 2);
+  }
+}
+
+.loginButton {
```
mattrosno, line 20: Update to `.login-button` to follow convention, or rename to `.button` because there's only one button in the component. That's probably cleanest.

services/web-app/components/PageNotFound/PageNotFound.js
```diff
@@ -0,0 +1,47 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [loading, isAuthenticated, user])
+
+  return (
+    <Grid className={styles.grid}>
+      <Column className={styles.column} sm={4} md={6} lg={6} xlg={6} max={6}>
+        <h1 className={styles.title}>Page not found.</h1>
+        {!isAuthorized && (
+          <>
+            <h1 className={styles.title}>Log in to view all pages.</h1>
```
mattrosno, line 33: Update to `<h2>` so we only have one h1 on the page (avoid accessibility violation.)

services/web-app/components/PageNotFound/PageNotFound.js
```diff
@@ -0,0 +1,47 @@
...
+  useEffect(() => {
+    if (!loading && isAuthenticated && isValidIbmEmail(user?.email ?? '')) {
+      setisAuthorized(true)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [loading, isAuthenticated, user])
+
+  return (
+    <Grid className={styles.grid}>
+      <Column className={styles.column} sm={4} md={6} lg={6} xlg={6} max={6}>
```
mattrosno, line 29: I'm not sure you need the additional `xlg={6} max={6}` because lg is 6?

services/web-app/components/PageNotFound/PageNotFound.module.scss
mattrosno, line 20: Can you add a comment saying that you're overriding carbon max-width and padding-right? You'll see that in our Sass as we need to override Carbon to help explain why we're setting those declarations.

services/web-app/components/PageNotFound/PageNotFound.module.scss
```diff
@@ -0,0 +1,24 @@
...
+.title {
+  @include type-style('display-01', true);
+}
+
+.grid {
+  margin-top: $spacing-10;
+  @include breakpoint-down(md) {
+    padding-right: calc($spacing-05 + var(--cds-grid-gutter) / 2);
+    padding-left: calc($spacing-05 + var(--cds-grid-gutter) / 2);
+  }
```
mattrosno, line 17: I think you can do this for simplicity, how we've handled it elsewhere.

```scss
include breakpoint-down(md) {
    padding-right: $spacing-05;
    padding-left: $spacing-05;
  }
```

mattrosno (changes requested):
Looks great! Some small suggestions.

mattrosno (approved):
@francinelucca I pushed a small cleanup commit and will merge after everything passes!

PR #277, opened 2022-02-23
 Flucca/chore/minor schema updates

francinelucca:
rename private to noIndex, move inherits property from assets to library
remove inheritable properties from required array


mattrosno (changes requested):
I'm game to do the schema changes in the PR if you'd like. IMO, these changes would technically be BREAKING and would require a major version bump, but since we haven't released anything yet, I wasn't overly concerned about semvar and these changes.

Side note: does the `perf` commit type make the most sense for these changes?

Let me know if you want to get this in first, then I can back changes out of https://github.com/carbon-design-system/carbon-platform/pull/271 and merge that second.

francinelucca:
@mattrosno I agree this would have to be a major version change (breaking) . I asked joe what's the commit msg syntax have to be so I can ammend the commit 'cause I'm not sure. That's why I have the do not merge for right now

mattrosno (approved):

PR #278, opened 2022-02-23
fix(web-app): redirect to previous page on local login

francinelucca:
move local auth logic further down because session.next is not being set locally since it's redirecting before it gets there


mattrosno (approved):
In Incognito, when directly to http://localhost:3000/assets/libraries, clicked log in icon, and it redirected me back to http://localhost:3000/assets/libraries after locally logging in. From what I remember from earlier today's behavior, this is now working as expected. LGTM

PR #288, opened 2022-02-25
Update service-config.prod.json

jharvey10:


service-config.prod.json
```diff
@@ -1,5 +1,10 @@
 {
-    "deployedServices": {
-
+  "deployedServices": {
+    "@carbon-platform/web-app": {
+      "deployedVersion": "0.1.56-beta.1"
```
francinelucca, line 4: can we build a new beta image so that login works?

jdharvey-ibm: yes!

mattrosno (approved):

francinelucca (approved):

alisonjoseph (approved):

PR #289, opened 2022-02-25
chore: make env-specific library lists

mattrosno:
Simplifies our library allowlists to set us up for prod-specific registering of libraries.

#### Testing / reviewing

Read the comments in that file to see how this works.

https://github.com/carbon-design-system/carbon-platform/blob/641c258ec1d18286a5f613ff3adb3cb54391702a/services/web-app/data/libraries.js

### DIFF
--- a/services/web-app/data/libraries.js
+++ b/services/web-app/data/libraries.js
@@ -38,10 +38,11 @@ const watsonMomentsGroup = {
 */
 
 /**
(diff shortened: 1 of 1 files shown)


francinelucca (approved):
LGTM

mattrosno:
@jdharvey-ibm thoughts on this to manage environment-specific data until we have a service in place?

PR #293, opened 2022-02-25
chore(web-app): clean up asset landing page and components

mattrosno:
Some Next.js cleanup.

#### Changelog

**New**

- v0.2 had missing content and bullets

**Changed**

- Renamed `PageHeaderLarge` component to `Hero` because as named, that seems like it's a larger variant of the existing page header and they don't really share much in common
- Added props to `Hero` so an image can be passed in and not hardcoded
- Hero component now shows the illustration at small breakpoint
- Moved all images needed by assets LP to the pages directory because I don't think Next.js needs all statically-imported images in the `/public` folder any more
- Moved the `AssetLandingPage` from its own file to the `index.js` because this component will be used once, in the index page, so it doesn't need to live separately like it is its own page. It's really just the content for that page.
- Update copy on the page so the text is directly in the HTML instead of being rendered as `{'strings'}`
- Add blur image placeholder to body images
- Update divs to semantic elements, best practices like only one h1 per page, etc.

**Removed**

- N/A

#### Testing / reviewing

Images blur in, missing copy is in, content is grid aligned.


alisonjoseph (dismissed):
🚀

andreancardona (approved):

PR #314, opened 2022-03-02
feat: added esc working on focus

### LINKED ISSUE #163: CatalogMultiselectFilter enhancements
### Browser

_No response_

### Description

There are accessibility issues with the new multiselect custom component, but also some UX issues that need addressed:

- [x] Close popover on escape key
- [x] Move focus to popover on open and give it the thin blue focus outline, like the normal multiselect component
- [x] Update state on close so the popover doesn't close each time you select a new item
- [x] Add motion to popover opening/closing (slide down like dropdown?)
- [x] Can we nest a tag in a multi-select? We're currently getting "button in button" accessibility warnings.

### Steps to reproduce

- Open the multi select and the popover does not have focus, pressing escape does not close the popover, and the popover should not close when you select an item

andreancardona:
Closes #163 

## Testing: 

-  Make sure you can close Popover using ESC
- Make sure focus  moves to popover on open
- Make sure query state is updated so that the popover remains open each time a new tag is selected


services/web-app/components/catalog-multislect-filter/catalog-multiselect-filter.js
```diff
@@ -33,6 +33,19 @@ const CatalogMultiselectFilter = ({
   const triggerRef = useRef()
   const contentRef = useRef()
 
+  useEffect(() => {
+    const handleEsc = (event) => {
+      if (event.keyCode === 27) {
+        setOpen(false)
+      }
+    }
+    window.addEventListener('keydown', handleEsc)
```
francinelucca, line 42: is it possible to use `useEventListener` here? see line 41: https://github.com/carbon-design-system/carbon-platform/blob/42ac07e96f91671adcaa66746cdc51c447b634c8/services/web-app/components/catalog-sort/catalog-sort.js#L41

andreancardona: Oo I'm moving this to it's own function.  let me know what you think

francinelucca (changes requested):
LGTM! behavior as expected and code looks good, just got question on double call to useFocus()

services/web-app/components/catalog-multislect-filter/catalog-multiselect-filter.js
```diff
@@ -12,26 +12,33 @@ import PropTypes from 'prop-types'
...
 }) => {
-  const [open, setOpen] = useState(false)
   const isLg = useMatchMedia(mediaQueries.lg)
   const isMd = useMatchMedia(mediaQueries.md)
   const triggerRef = useRef()
   const contentRef = useRef()
+  const popoverRef = useRef(null)
+  const focus = useFocus()
+
+  useFocus()
```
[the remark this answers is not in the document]
andreancardona: nada - removed it, thank you for the catch :)

francinelucca (dismissed):
🎉

mattrosno:
It looks like #260 needs merged into main, and then main into this branch so `use-query-state.js` works with the latest code, then should be ready for review? @andreancardona @francinelucca

francinelucca:
> It looks like #260 needs merged into main, and then main into this branch so `use-query-state.js` works with the latest code, then should be ready for review? @andreancardona @francinelucca

Yeah let's merge refactor in first since that also solves for other problems in query state that will allow this to work as expected once in

mattrosno:
@andreancardona @francinelucca not sure if this should be part of this issue, but we need to set a height on the popover content because when its parent is position sticky, and when the popover content is tall, you can't get to the bottom of the popover content.

We could achieve this by setting overflow scroll (whatever is needed for smooth scroll iOS in CSS) and then a max-height that's 100vh minus whatever's above it.

It'd be nice to not need a new dependency for this, but this library is popular to get around cross browser mobile quirks.

https://www.npmjs.com/package/react-div-100vh

This should probably be documented in a new issue.

image

andreancardona:
> @andreancardona @francinelucca not sure if this should be part of this issue, but we need to set a height on the popover content because when its parent is position sticky, and when the popover content is tall, you can't get to the bottom of the popover content.
> 
> 
> 
> 
> This should probably be documented in a new issue.
> 
> image

Going to add this to a new issue!

services/web-app/components/catalog-multislect-filter/catalog-multiselect-filter.js
```diff
@@ -23,11 +25,18 @@ const CatalogMultiselectFilter = ({
   className: customClassName,
   onFilter
 }) => {
-  const [open, setOpen] = useState(false)
   const isLg = useMatchMedia(mediaQueries.lg)
   const isMd = useMatchMedia(mediaQueries.md)
   const triggerRef = useRef()
   const contentRef = useRef()
+  const popoverRef = useRef(null)
+  // const focus = useFocus()
```
francinelucca, line 33:
```suggestion
```

services/web-app/components/catalog-multislect-filter/catalog-multiselect-filter.js
```diff
@@ -90,7 +102,8 @@ const CatalogMultiselectFilter = ({
         </span>
       </button>
       <PopoverContent className={styles.content} ref={contentRef}>
-        <Column span={columns}>
+        {/* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */}
```
francinelucca, line 105: can we add a comment as to why this is necessary?

services/web-app/utils/use-query-state.js
```diff
@@ -6,7 +6,7 @@
  */
 
 import { useRouter } from 'next/router'
-import { useCallback, useEffect, useMemo } from 'react'
+import { useCallback, useEffect, useState } from 'react'
```
francinelucca, line 9: can you pull from main so all this gets dismissed?

francinelucca (changes requested):

alisonjoseph (approved):
Working as expected! 🥳

francinelucca (approved):

PR #315, opened 2022-03-02
Validate correct page number when affecting changes occur

### LINKED ISSUE #310: web-app(catalog): broken pagination validation
### Browser

_No response_

### Description

We're missing some navigation validations for the pagination component in the catalog which makes it possible to get stuck with no content (on search or page change) when there should be content

### Steps to reproduce

1 - Select an items per page amount that allows you to navigate to second page
2 - Go to second page
3 -  Select an items per page amount that makes all content fit in first page 

You are now stuck with no content

------
1 - Select an items per page amount that allows you to navigate to second page
2 - Go to second page
3 -  Perform a search that returns a small subset of items

You are now stuck with no content

-------
1 - manually modify page number in browser query to an invalid page number

You are now stuck with no content

francinelucca:
Closes #310 

- [x] Check re-render behavior once #260 gets merged in (specifically page2 -> search)


francinelucca:
Need to get in #260 first due to unstable use-query-state behavior regarding useState()

services/web-app/components/catalog/catalog.js
```diff
@@ -348,6 +348,14 @@ function Catalog({ collection, data, type, filter: defaultFilter = {}, glob = {}
     setPossibleFilterValues(getFilters({ collection, type }))
   }, [collection, type])
 
+  useEffect(() => {
+    const resultsCount = filteredAssets.length
+    const maxPageNumber = Math.max(Math.ceil(resultsCount / pageSize), 1)
+    if (page > maxPageNumber) {
```
mattrosno, line 354: Might just want to add one more check if page isn't an integer, set it to 1. E.g. if you manually put `&page=asdf` in the URL.

mattrosno (changes requested):

services/web-app/components/catalog/catalog.js
francinelucca: I didn't add it because in that case, useQueryState will dismiss the invalid value after calling the supplied validation function (line 219) 
image
so it will return the last valid value, by the time it gets here the input has already been cleared.
Do you still want me to add a check for redundancy?

mattrosno: The page held up just fine with `&page=asdf`, so we're good! Thanks for explaining how `useQueryState` is handling this.

mattrosno (dismissed):

services/web-app/components/catalog/catalog.js
jdharvey-ibm: There's a few edge cases here for things that can be coerced to numbers. All of these will pass the `!isNaN` check, but might not be returning something useful:

```js
!isNaN(false)
!isNaN(null)
!isNaN(true)
!isNaN(1.123e9)
!isNaN('')
```

`parseInt` might be a better validation option, since I'm not sure if `parseNumbers: true` tries both `parseInt` and `parseFloat`.

francinelucca: Done!

mattrosno (approved):

alisonjoseph (approved):

PR #329, opened 2022-03-04
chore(web-app): save page, pageSize, and query to localStorage

### LINKED ISSUE #313: Persist filter, search, and pagination to local storage like sort and view

francinelucca:
Closes #313


francinelucca:
Need to get #260  in first

services/web-app/utils/use-query-state.js
```diff
@@ -95,6 +95,14 @@ export const useQueryState = (
     if (saveToStorage && value) {
       localStorage.setItem(`${router.pathname}:${key}`, value)
```
mattrosno, line 96: Is this line needed given line 100 below?

francinelucca: it's not! you're right. I'm going to remove these changes anyway since I already took care of this over on #260

francinelucca:
closing, might reopen once we get resolution on #313

PR #334, opened 2022-03-07
docs(web-app): catalog filter comments

mattrosno:
@francinelucca as I was reviewing https://github.com/carbon-design-system/carbon-platform/pull/260, I found myself wanting to add comments to previous logic not touched in that PR, but also new logic in that PR. Instead of a PR review, what about these changes if you want to incorporate them? In addition to adding comments, I move a couple functions out of the `Catalog` component to make them pure functions. It was throwing me off that `getFilteredAssets` and `filterPropertyHasValidValue` were included in the middle of all the state variables.

### DIFF
--- a/services/web-app/components/catalog/catalog.js
+++ b/services/web-app/components/catalog/catalog.js
@@ -28,12 +28,26 @@ import useQueryState from '@/utils/use-query-state'
 
 import styles from './catalog.module.scss'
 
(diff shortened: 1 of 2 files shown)


francinelucca (approved):
LGTM!, thanks for helping me clean that up hehe :tada @mattrosno

PR #338, opened 2022-03-08
fix: broken link styling asset lp

mattrosno:
The three links on the asset landing page are too small. Upon further review, I found that we were nesting anchor elements. Easy mistake, given how we are using the Carbon `Link` component here, no the Next `Link` component that requires a child anchor element.

#### Changelog

**New**

- NA

**Changed**

- Link styling on asset landing page

**Removed**

- NA

#### Testing / reviewing

Links in asset landing page are now the same size as their surrounding body copy.

### DIFF
--- a/services/web-app/pages/assets/index/index.module.scss
+++ b/services/web-app/pages/assets/index/index.module.scss
@@ -37,6 +37,7 @@
 .link {
   // override carbon
   font-size: inherit;
+  line-height: inherit;
 }
 
 /**
--- a/services/web-app/pages/assets/index/index.module.scss
+++ b/services/web-app/pages/assets/index/index.module.scss
@@ -35,7 +35,8 @@
 }
 
 .link {
-  @include type-style('body-long-02');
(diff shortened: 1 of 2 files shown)


services/web-app/pages/assets/index/index.module.scss
```diff
@@ -34,6 +34,10 @@
   font-weight: 600;
 }
 
+.link {
+  @include type-style('body-long-02');
```
alisonjoseph, line 38: Do we need the token here, could we not just do `font-size: inherit` so it picks up on the token from the paragraph and matches the size? That way I think we would be covered in the future if we wanted to add a link to any of the larger type sections.

mattrosno: That works too. Updating.

alisonjoseph (approved):

francinelucca:
Reviewing now

francinelucca (approved):
LGTM!

Issue #296, opened 2022-02-28
Add mocked library and asset data for tests


mattrosno:
In our monorepo, if we could stub out test libraries and assets that hit edge cases like:

- [ ] Missing schema keys
- [ ] Inheritance examples
- [ ] Branch and version (tag) data fetching

Then we can either use that data in Storybook stories for E2E testing and VRT with Percy, and we could also use it for unit tests so we don't have to mock all requests and responses.

PR #341, opened 2022-03-08
feat(web-app): add library and asset validation

francinelucca:
Relates to #296 

Adds asset and library validation: skip incomplete assets or libraries

Note: I'm open to moving some of this stuff around if there's a feel they're in the wrong place, wasn't really sure where to place them. Same thing with var/function naming.

#### Testing / reviewing

Pull changes and confirm app still works as expected and same content still gets loaded.

Additional: if you want to test with invalid assets you can:
- Libraries: Add something like `delete library.id` or `library.demoLinks = [{test: 'test'}]` on `line 134` of` services/web-app/lib/github.js`, reload and you should see different content and some warnings in your project terminal
- Assets: Add something like `delete asset.content.id` or `asset.content.demoLinks = [{test: 'test'}]` right before line 178
 of `services/web-app/lib/github.js`, reload and you should see different content and some warnings in your project terminal


jharvey10 (changes requested):

francinelucca:
@mattrosno updated, no longer skipping assets w.o description

jharvey10 (dismissed):

services/web-app/lib/github.js
```diff
@@ -153,10 +201,18 @@ export const getLibraryData = async (params = {}) => {
...
     ? assets.filter((asset) => getSlug(asset.content) === libraryParams.asset)
     : assets
 
+  filteredAssets = filteredAssets.filter((asset) => {
+    const isValidAsset = validateAsset(asset.content, library)
+    if (libraryParams.asset) {
+      return isValidAsset && getSlug(asset.content) === libraryParams.asset
+    }
+    return isValidAsset
+  })
```
mattrosno, line 217: Do we need that first filteredAssets? I think this can be simplified to:

```suggestion
  const filteredAssets = assets.filter((asset) => {
    const isValidAsset = validateAsset(asset.content, library)
    if (libraryParams.asset) {
      return isValidAsset && getSlug(asset.content) === libraryParams.asset
    }
    return isValidAsset
  })
```

mattrosno (changes requested):
One small question. Also, I pushed a commit updating some JS comments.

jharvey10 (dismissed):

services/web-app/lib/github.js
francinelucca: You're right! fixed!

mattrosno (approved):

andreancardona (approved):
looks good! 🎉

PR #348, opened 2022-03-09
non-Canonical Framework Collapsing

### LINKED ISSUE #347: Catalog framework collapsing doesn't work when canonical library's asset is `noIndex`
### Browser

_No response_

### Description

As shown in the status table, network diagram component only exists in angular and react.

https://github.com/carbon-design-system/carbon-charts#component-status

Because charts is set up as a group where the framework libraries inherit the vanilla library, we still want network diagram in the vanilla library, but not shown in the catalog.

https://github.com/carbon-design-system/carbon-charts/blob/0c374c1cbd7b9b77fcef59b8e967590df287d50d/packages/core/carbon.yml#L373

And then network diagram also exists in angular (https://github.com/carbon-design-system/carbon-charts/blob/cdd55646b2812bae8d85932d38b1ed464f5d67a3/packages/angular/carbon.yml#L194) and react (https://github.com/carbon-design-system/carbon-charts/blob/cdd55646b2812bae8d85932d38b1ed464f5d67a3/packages/react/carbon.yml#L194).

This means that if no framework is set, it tries to render the vanilla network diagram, but that's `noIndex: true`, so network diagram doesn't show up by default.

image

But if you filter by framework, framework collapsing doesn't happen anymore, so you see it.

image

**Options**

1. Because carbon charts react has all components, we could switch canonical from vanilla to react.
2. Instead of defining one canonical library per group, we could make it an array to better pick a framework when no framework is in the filter.
3. Remove the `noIndex` on the vanilla network diagram so it'll show "design only +2"

### Steps to reproduce

See above.

francinelucca:
Closes #347 

Collapses assets even if they're not in the canonical library (see network graph from carbon charts)

Testing
Validate that search, list view and grid view work as expected, that all carbon charts show vanilla-first framework collapsing and network graph shows angular-first framework collapsing

### DIFF
--- a/services/web-app/components/catalog/catalog.js
+++ b/services/web-app/components/catalog/catalog.js
@@ -132,21 +132,34 @@ function Catalog({ collection, data, type, filter: defaultFilter = {}, glob = {}
 
   useEffect(() => {
     const skippedAssets = []
-    const assetsWithAppliedFilter = assets
-      .sort(assetSortComparator(sort))
-      .filter((asset) => {
-        if (assetIsInFilter(asset, filter)) {
-          if (collapseAssetGroups(asset, filter)) {
-            const isCanonicalAsset = isCanonicalLibAsset(asset)
-            if (!isCanonicalAsset) skippedAssets.push(asset)
-            return isCanonicalAsset
-          }
-          return true
-        } else {
-          return false
+    const assetsWithAppliedFilter = assets.filter((asset) => {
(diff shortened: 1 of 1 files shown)


mattrosno (approved):
Looks great! Result counts still look good. I tried searching, sorting, filtering. Cool to see the Angular default in this example.

image

alisonjoseph (approved):
🚀 LGTM!

PR #349, opened 2022-03-10
fix: align nav to grid

### LINKED ISSUE #281: UI shell nav items should align with side nav
### Browser

_No response_

### Description

Per @jeanservaas via @andreancardona the left edge of the Assets global nav item should align with the side nav container. The UI shell is not grid aware, so if we do this, we'll have to override Carbon.

### Steps to reproduce

image

alisonjoseph:
Closes #281
Screen Shot 2022-03-10 at 8 52 01 AM

#### Changelog
- Removes border to the right of the title
- Aligns top nav to grid

#### Testing / reviewing
Check that top nav aligns with grid at all breakpoints

### DIFF
--- a/services/web-app/layouts/layout/layout.module.scss
+++ b/services/web-app/layouts/layout/layout.module.scss
@@ -36,8 +36,6 @@
 .header-name {
   @include breakpoint(lg) {
     position: absolute;
-    // match width of side nav
-    width: 16rem;
   }
 }
 
--- a/services/web-app/layouts/layout/layout.module.scss
+++ b/services/web-app/layouts/layout/layout.module.scss
@@ -11,9 +11,11 @@
   background: $background;
 }
 
-:global(.cds--header__nav::before) {
-  // override Carbon, hide divider line
-  display: none;
+.header {
+  :global(.cds--header__nav::before) {
+    // override Carbon, hide divider line
+    display: none;
+  }
 }
 
 .main {
--- a/services/web-app/layouts/layout/layout.module.scss
+++ b/services/web-app/layouts/layout/layout.module.scss
@@ -41,13 +41,14 @@
   }
 }
 
-.header-navigation {
+.header-grid {
   @include breakpoint(lg) {
-    //align nav to grid
-    margin-left: 24.3%;
+    width: 100%;
+    height: 100%;
+    margin: 0;
   }
+}
 
-  @include breakpoint(max) {
-    margin-left: 390px;
-  }
+.header-navigation {
+  margin-left: -$spacing-05;
 }
--- a/services/web-app/layouts/layout/layout.module.scss
+++ b/services/web-app/layouts/layout/layout.module.scss
@@ -11,6 +11,11 @@
   background: $background;
 }
 
+:global(.cds--header__nav::before) {
+  // override Carbon, hide divider line
+  display: none;
+}
+
 .main {
   // position for PageBreadcrumb
   position: relative;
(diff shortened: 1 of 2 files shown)


francinelucca (dismissed):
LGTM! right line on top bar is gone and nav options aligned to grid on all breakpoints 🎉 
image

services/web-app/layouts/layout/layout.module.scss
```diff
@@ -11,6 +11,11 @@
   background: $background;
 }
 
+:global(.cds--header__nav::before) {
+  // override Carbon, hide divider line
+  display: none;
+}
```
mattrosno, line 17: Small detail, but I think it's a good practice for any `:global()` to still be scoped to a moduled class name. For example, in `catalog-filters.module.scss`, we override `.cds--tag` but only inside the `.container` class, which gets locally scoped.

I know header nav is global and we only expect there to be one on the page, but to play it safe you could:

```scss
.header {
  :global(.cds--header__nav::before) {
    // override Carbon, hide divider line
    display: none;
  }
}
```

And `<Header aria-label="Carbon Design System" className={styles.header}>`.

To locally scope that Carbon override.

mattrosno (changes requested):

services/web-app/layouts/layout/layout.module.scss
alisonjoseph: Updated!

francinelucca (dismissed):

services/web-app/layouts/layout/layout.module.scss
```diff
@@ -27,3 +34,20 @@
...
+.header-name {
+  @include breakpoint(lg) {
+    position: absolute;
+  }
+}
+
+.header-navigation {
+  @include breakpoint(lg) {
+    //align nav to grid
+    margin-left: 24.3%;
```
mattrosno, line 47: I'd approach this differently, so we don't have arbitrary styling values like `24.3%`.

With

```scss
.header-name {
  @include breakpoint(lg) {
    position: absolute;
  }
}

.header-grid {
  @include breakpoint(lg) {
    width: 100%;
    height: 100%;
    margin: 0;
  }
}

.header-navigation {
  margin-left: -$spacing-05;
}
```

We can properly use the `Grid` and `Column` component:

```jsx
            <Header aria-label="Carbon Design System" className={styles.header}>
              <SkipToContent />
              <HeaderMenuButton
                aria-label="Open menu"
                onClick={onClickSideNavExpand}
                isActive={isSideNavExpanded}
              />
              <div className={styles.headerName}>
                <Link href="/assets">
                  <a className="cds--header__name">Carbon Design System</a>
                </Link>
              </div>
              <Grid condensed className={styles.headerGrid}>
                <Column sm={0} lg={{ span: 8, offset: 4 }}>
                  <HeaderNavigation
                    aria-label="Main navigation"
                    className={styles.headerNavigation}
                  >
                    {globalNavData.map((data) => (
                      <HeaderMenuItem
                        key={data.path}
                        isCurrentPage={router.pathname.startsWith(data.path)}
                        href={data.path}
                        element={NextLink}
                      >
                        {data.title}
                      </HeaderMenuItem>
                    ))}
                  </HeaderNavigation>
                </Column>
              </Grid>
            </Header>
```

mattrosno (changes requested):
Can we use Carbon's grid for this instead of setting arbitrary values? How I'd approach it below.

image

mattrosno (changes requested):

mattrosno (approved):
💯

jharvey10 (approved):

PR #390, opened 2022-03-15
wip(web-app): log message api and call from middleware

### LINKED ISSUE #245: Web App: Log incoming page requests with the logging service
- Can assist debugging to see web page traffic before/after incidents

Phase 1: create an internal-only logging api in the next app which takes a log message as input and sends it to the messaging service. Call that from the existing middleware to log incoming request data.
Phase 2: create a minimal express server to receive incoming requests and proxy them to the actual nextjs application. This will allow both the request headers and the response code to be retrieved and logged. Remove the logging api in favor of the express-based logging.

francinelucca:
Closes #245 
Sample Log Message (I believe IP is undefined because of running locally): 
image

#### Testing / reviewing

Run with changes and see console when navigating through the app, page requests should be logged


services/web-app/pages/_middleware.js
```diff
@@ -32,12 +32,26 @@ function exitWith404(req) {
 }
 
 export async function middleware(req) {
+  const protocol =
+    getRunMode === RunMode.Prod || process.env.RUNNING_SECURELY === '1' ? 'https' : 'http'
+
+  // log page request
+  fetch(`${protocol}://localhost:${req.nextUrl.port}/api/log-message`, {
```
jdharvey-ibm, line 39: Do you think we should have a helper function outside of this middleware which takes a request and generates a loopback fetch URL?

jdharvey-ibm: Or maybe a wrapper around sending loopback requests altogether. Basically something that would handle all of the stuff we need to hand-write for each of these fetch requests.

services/web-app/pages/_middleware.js
```diff
@@ -32,12 +32,26 @@ function exitWith404(req) {
...
+    body: JSON.stringify({
+      logMessage: `[${new Date().toISOString()}] "${req.method} ${req.nextUrl.pathname}" "${
+        req.ua.ua
+      }" "${req.ip}"`
+    }),
+    headers: {
+      'Content-Type': 'application/json'
+    }
+  })
+
```
jdharvey-ibm, line 50: This middleware function is getting a bit lengthy and does a few different things. Would you mind refactoring it to delegate the individual pieces of the middleware to their own functions which are called by `middleware()`? Here's a starting point that I had locally when I tried to do the auth stuff. You could add another function to the `middleware` function to perform the logging.

```js
/*
 *
 */

import { getRunMode, RunMode } from '@carbon-platform/api/runtime'
import cookie from 'cookie'
import { NextResponse } from 'next/server'

import { isValidIbmEmail } from '@/utils/string'

class NotAuthorizedError extends Error {}

/**
 * Creates request options for retrieveUser api call including headers and agent if necessary
 *
 * @param {NextRequest} req getServerSideProps request object
 * @returns {RequestInit} request options
 */
function getRequestOptions(req) {
  const sessionCookie = req.cookies['connect.sid']
  const cookies = sessionCookie ? cookie.serialize('connect.sid', sessionCookie) : ''
  return {
    headers: { cookie: cookies }
  }
}

function exitWith404(req) {
  const url = req.nextUrl.clone()
  url.pathname = '/404'
  return NextResponse.rewrite(url)
}

async function applyAuthMiddleware(req) {
  if (req.nextUrl.pathname === '/404' || req.nextUrl.pathname.startsWith('/api/')) {
    return
  }

  const protocol =
    getRunMode === RunMode.Prod || process.env.RUNNING_SECURELY === '1' ? 'https' : 'http'
  const userResponse = await fetch(
    `${protocol}://localhost:${req.nextUrl.port}/api/user`,
    getRequestOptions(req)
  )

  // Guard - non-200 user api response
  if (!userResponse.ok) {
    throw new NotAuthorizedError()
  }

  const user = await userResponse.json()

  // Guard - no valid user
  if (!user?.email) {
    throw new NotAuthorizedError()
  }

  // Guard - not a valid IBMer
  if (!isValidIbmEmail(user.email)) {
    throw new NotAuthorizedError()
  }
}

export async function middleware(req) {
  try {
    await applyAuthMiddleware(req)
  } catch (err) {
    return exitWith404(req)
  }

  return NextResponse.next()
}
```

services/web-app/pages/api/log-message.js
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright IBM Corp. 2022, 2022
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import { Logging } from '@carbon-platform/api/logging'
+
+const logging = new Logging('web-app', 'log-message-api')
```
jdharvey-ibm, line 9: Let's make this API very specific to logging incoming requests. Maybe name it `log-request.js` and change this component in the constructor here to `log-request-api`.

jharvey10 (changes requested):

services/web-app/pages/_middleware.js
francinelucca: added services/web-app/utils/sendLocalRequest.js

services/web-app/pages/_middleware.js
francinelucca: Thanks for the file! incorporated these changes and added call `logIncomingRequest()` to `middleware`

services/web-app/pages/api/log-message.js
francinelucca: Done!

services/web-app/pages/_middleware.js
```diff
@@ -5,58 +5,59 @@
...
 function exitWith404(req) {
   const url = req.nextUrl.clone()
   url.pathname = '/404'
   return NextResponse.rewrite(url)
 }
 
-export async function middleware(req) {
+function logIncomingRequest(req) {
+  sendLocalRequest(req, '/api/log-request', false, 'POST', {
+    logMessage: `[${new Date().toISOString()}] "${req.method} ${req.nextUrl.pathname}" "${
```
jdharvey-ibm, line 23: The Logging service will automatically add a date to the request for you.

```suggestion
    logMessage: `"${req.method} ${req.nextUrl.pathname}" "${
```

services/web-app/pages/_middleware.js
```diff
@@ -5,58 +5,59 @@
...
+      req.ua.ua
+    }" "${req.ip}"`
```
jdharvey-ibm, line 25: Do we have access to the remote port from which the connection came? If so, it'd be good to add that in here as well.

services/web-app/pages/api/log-request.js
```diff
@@ -0,0 +1,28 @@
...
+    return
+  }
+
+  if (!req.body.logMessage) {
+    console.log('bad request no logmessage')
+    res.status(400).send('Bad request (no logMessage)')
+    return
+  }
+
+  // TODO: block external requests
```
jdharvey-ibm, line 24: Just adding some of the ideas here for good measure:

- Compare incoming request hostname to "localhost"
- Compare incoming request IP address to our local IP

services/web-app/utils/sendLocalRequest.js
```diff
@@ -0,0 +1,60 @@
...
+    headers: {
+      'Content-Type': 'application/json',
+      cookie: forwardAuthCookie ? getAuthCookie(req) : ''
+    },
+    ...requestOptions
+  }
+
+  if (method !== 'GET') {
+    reqOptions.body = JSON.stringify(body)
+  }
```
jdharvey-ibm, line 57: It might be easier to default `body` to null. If it's found, then set it in the reqOptions, if not, don't set the variable. There's a bunch of http request types that might need a body, so rather than enumerating them all here, just delegate that to the caller.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods

(If you click on each one, it'll tell you whether it needs a request body)

francinelucca: Done!

services/web-app/pages/api/log-request.js
francinelucca: Opted for env variable approach

services/web-app/pages/_middleware.js
francinelucca: Going no-port phase 1

jdharvey-ibm: We'll revisit this under #426 and merge with what we can get easily for now.

services/web-app/config/constants.js
```diff
@@ -4,14 +4,18 @@
...
 
-export const ALLOWED_REFERERS = [
+const ALLOWED_REFERERS = [
   { protocol: 'https:', domain: '.carbondesignsystem.com' },
   { protocol: 'https:', domain: 'web-app.j73b4w218e4.us-south.codeengine.appdomain.cloud' },
   { protocol: 'http:', domain: 'localhost' },
   { protocol: 'https:', domain: 'localhost' }
 ]
+
+const { CARBON_INTERNAL_API_SECRET } = loadEnvVars({ CARBON_INTERNAL_API_SECRET: 'DEV_SECRET' })
```
jdharvey-ibm, line 19: Can you add a little blurb to the web app docs to call out this envvar? 😁

services/web-app/pages/_middleware.js
```diff
@@ -5,58 +5,59 @@
...
   if (!isValidIbmEmail(user.email)) {
+    throw new NotAuthorizedError()
+  }
+
+  return NextResponse.next()
+}
+
+export async function middleware(req) {
+  try {
+    logIncomingRequest(req)
```
jdharvey-ibm, line 58:
```suggestion
  logIncomingRequest(req)
  try {
```

the log should be safe to call on its own and if it fails, it shouldn't result in a 404 to the client.

services/web-app/pages/api/log-request.js
```diff
@@ -0,0 +1,36 @@
...
+    res.status(400).send('Bad request (no logMessage)')
+    return
+  }
+
+  if (!internalApiSecret || internalApiSecret !== CARBON_INTERNAL_API_SECRET) {
+    logging.warn(`unauthorized access from ${req.socket.remoteAddress}`)
+    res.status(401).send('Bad request (no internalApiSecret)')
+    return
+  }
+
```
jdharvey-ibm, line 31: I'm loving the use of guard clauses 😁

services/web-app/pages/api/log-request.js
```diff
@@ -0,0 +1,36 @@
...
+  const { logMessage, internalApiSecret } = req.body
+
+  if (!logMessage) {
+    res.status(400).send('Bad request (no logMessage)')
+    return
+  }
+
+  if (!internalApiSecret || internalApiSecret !== CARBON_INTERNAL_API_SECRET) {
+    logging.warn(`unauthorized access from ${req.socket.remoteAddress}`)
```
jdharvey-ibm, line 27:
```suggestion
```

Let's just silently ignore these so external users can't flood the logs by submitting malformed requests.

services/web-app/pages/api/log-request.js
```diff
@@ -0,0 +1,36 @@
...
+    res.status(401).send('Bad request (no internalApiSecret)')
```
jdharvey-ibm, line 28:
```suggestion
    res.status(401).send('Unauthorized.')
```

services/web-app/utils/sendLocalRequest.js
```diff
@@ -0,0 +1,60 @@
...
+ * @param {NextRequest} req - Original request object that came through the middleware
+ * or getServerSideProps.context
+ * @param {string} path - Relative path to request
+ * @param {boolean} forwardAuthCookie - Should auth cookie be forwarded to local request
+ * @param {string} method - Request method to use, default to 'GET'
+ * @param {Object} body - Body object to send to fetch command
+ * @param {Object} requestOptions - Additional request options to send to the fetch command
+ * @returns {Promise<Response>} Promise return from fetch()
+ */
+export default function sendLocalRequest(
```
jdharvey-ibm, line 33: There's a fetch in `retrieveUser.js` that could benefit from this function. We should definitely update that one either in this PR or in a separate one.

There's also a couple in `auth.js`, but those are so simple it might not be worth it.

jharvey10 (changes requested):

services/web-app/config/constants.js
francinelucca: Done!

services/web-app/utils/sendLocalRequest.js
francinelucca: Now using in `retrieveUser.js`, `auth.js` doesn't have the absolute url constraint so I don't think it's necessary

jharvey10 (approved):

jdharvey-ibm:
Merging this in with just one review to get it in the nightly. Francine and I have both tested this pretty extensively in our local environments. I'll give it additional validation on the test env once we deploy it there.

PR #394, opened 2022-03-16
fix(schemas): asset description to optional

mattrosno:
In talking with @jeanservaas, we determined that asset descriptions don't have to be required. We don't want to discard indexed assets if they don't have descriptions yet. Instead, we'll show them with no descriptions, and then actively add descriptions when we see them missing in the web app.

Nothing needs changed in the web app.

Schema validation isn't enforcing required descriptions because of inheritance.

Do we need to update anything else?


mattrosno:
@francinelucca should we merge this in before https://github.com/carbon-design-system/carbon-platform/pull/341?

francinelucca:
Yes , let's do it, this LGTM

francinelucca (approved):

alisonjoseph (approved):

PR #397, opened 2022-03-17
fix(web-app): catalog sticky controls

### LINKED ISSUE #317: The search and sort bars aren't sticky any more
### Browser

_No response_

### Description

These used to be sticky on the catalog pages as you scrolled. I wonder when this stopped working...

### Steps to reproduce

image

mattrosno:
Closes #317 

When we built the asset landing page, we put `overflow:hidden` on the layout body to prevent horizontal scroll bars due to pseudo-element backgrounds and background SVG images (in the hero component.)

I just learned that `position:sticky` doesn't work if any parent is `overflow:hidden`. This PR removes the layout `overflow:hidden` and applies new styling in the assets landing page.

#### Changelog

**New**

- N/A

**Changed**

- Assets landing page highlights background pseudo element
- Hero background SVG positioning

**Removed**

- Layout `overflow:hidden`

#### Testing / reviewing

- Search and sort are sticky in asset catalogs again
- At all breakpoints, there's no horizontal scrolling in the assets landing page
- Hero background SVG positioning still looks good (this SVG will get updated to a new design soon - it's slightly different at extreme max viewport but I think that's okay until we receive those updated designs)
- Assets landing page highlights section has full width white background

### DIFF
--- a/services/web-app/components/hero/hero.module.scss
+++ b/services/web-app/components/hero/hero.module.scss
@@ -6,6 +6,7 @@
 //
 
 $hero-height: $spacing-13 * 3.475;
+$illo-mobile-height: $spacing-08 + $spacing-12;
 $illo-original-width: 1260px;
 $illo-original-height: 1078px;
 
@@ -36,7 +37,7 @@ $illo-original-height: 1078px;
 .column {
   position: relative;
   z-index: 1;
-  padding: $spacing-08 $spacing-05 ($spacing-08 + $spacing-12);
+  padding: $spacing-08 $spacing-05 $illo-mobile-height;
   @include breakpoint(md) {
     padding: ($spacing-09 + $spacing-03) 0 $spacing-05;
   }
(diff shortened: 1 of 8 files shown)


services/web-app/components/hero/hero.module.scss
```diff
@@ -56,18 +57,36 @@ $illo-original-height: 1078px;
...
   width: 50%;
   height: $hero-height;
   @include breakpoint(xlg) {
     // -18vw chosen to horizontally center the illo in white space at ultra max viewport
     right: -18vw;
     width: $hero-height * ($illo-original-width / $illo-original-height);
   }
+  @include breakpoint(max) {
+    // match xlg position
+    right: -285px;
```
andreancardona, line 88: not that it matters too much but could we so some version of this with tokens?  this below equates to 17.5 rem vs.  285px. (17.813 rem) or does it have to be exactly 285px.  

I know this is a very small details -  was just curious 😊

alisonjoseph (changes requested):
I'm still seeing a horizontal scrollbar in `/assets/components` page and asset landing page `assets`

mattrosno:
I'll take a look at scroll bars.

services/web-app/components/hero/hero.module.scss
mattrosno: @andreancardona this is positioning the background SVG so tokens don't make a lot of sense here. We're taking the generic SVG, and scaling and positioning it to match the spec. Because I couldn't use tokens here, I added CSS comments to explain the arbitrary values.

mattrosno:
I can recreate the overflow issue in MacOS Chrome with the "always" setting.

image

mattrosno:
@alisonjoseph you can see what I did here: https://github.com/carbon-design-system/carbon-platform/pull/397/commits/87167d5a7059cb205d92a80c8d1ed736c222a5cc

francinelucca (approved):
- [x]  sticky sort + search 
- [x] No horizontal bar at all breakpoints
- [x] full width (`calc(100vw - var(--scrollbar-width))`) + white background in asset landing highlight 
- [x] Hero LGTM

Approved

alisonjoseph (approved):
Awesome, looks good Matt!

PR #400, opened 2022-03-18
feat(web-app): breadcrumb and page header enhancements

### LINKED ISSUE #368: Color code asset types
**Design**

- [x] Spec changes @kingtraceyj 

**Development**

- [x] Type tags in collections and libraries (no catalogs because those are single-type)
- [x] Page header colors

### LINKED ISSUE #387: Breadcrumb takes you out to global asset catalog (its type)

mattrosno:
Closes #387
Closes #368

#### Changelog

**New**

- Added an override to global selected nav item so the bottom border isn't indented two pixels in each side (his is so it'll align with page tabs when page tabs become sticky)
- Adds space between 404 page content and footer
- Catalog items now have colored type tags
  - TypeTag component for this
- Page header pictogram in libraries and library detail pages (custom icon because it's not in our pictograms package)
- Global Sass variables file so we can use page header height in both page breadcrumb and page header styling
- `/data/nav-data.js` so we can define a plural string per asset type and also pictograms and colors (which use Carbon's colors JS package)

**Changed**

- Breadcrumb items in asset detail page to go out to catalogs not library per spec
- Page header colors according to spec
- Pictogram in page header is aligned to right of viewport not content
- Moved global Carbon overrides to `/styles/vender/_overrides.scss`

**Removed**

- NA

#### Testing / reviewing

- Default page header background is now black
- Asset catalogs have background colors
- Catalog items have a type tag next to framework icon


mattrosno:
@kingtraceyj some screenshots to review. I can't remember if you wanted to only show the type tags in collections not catalogs, but I really don't mind it there too. Yes it's a little repetitive, but once you've scrolled down and no longer see the purple page header, it's nice to see the purple pattern tags so you know what type you're viewing.

image
image
image
image
image
image
image
image

kingtraceyj:
Looking good!

For the time being, let's keep the asset type tag only in collections if possible. 

A few more changes:

- Template banner icon should be `dashboard`

image

- Library banner icon should be `backup-03`

image


- tag text should be the color-70

image

mattrosno:
Thanks @kingtraceyj! I'll update. Can you share the SVG for backup 03 pictogram? I don't see it in our library: https://www.carbondesignsystem.com/guidelines/pictograms/library

kingtraceyj:
@mattrosno 
[Backup-03.svg.zip](https://github.com/carbon-design-system/carbon-platform/files/8316150/Backup-03.svg.zip)

kingtraceyj:
one other note:

the banner icon should always be 32 px away from the edge and stick there.

currently the icon is sticking to the edge of the content area:
image

should be 32 from edge:
image

kingtraceyj:
yeah. it should stick to the right.

mattrosno:
@kingtraceyj @jeanservaas 

>For the time being, let's keep the asset type tag only in collections if possible.

Are you sure you only want to show the type tags if there is the potential for mixed types (collections, library assets table)?

After you've scroll down and no longer see the page header background color, it's a nice confirmation of what you're seeing.

image

Especially on mobile when there's no side nave to orient you.

image

I don't think they add too much visual noise. Just making sure we want to remove these in the type catalogs before I add logic to do so.

jeanservaas:
@mattrosno at the larger breakpoints this is really repetitive and busy... we're already at the bitter edge of what those cards can contain and several people we've played some things back to including eliad have mentioned that they felt overwhelmed by the amount of info on the cards.

I agree with the situation in mobile without the side nav though... could we adaptively add the tags there only? Just seems really unnecessary at the larger breakpoints.

mattrosno:
@jeanservaas thanks. If that's the case, let's keep it simple and not include them in the catalogs (but still in collections and library assets table when we have that).

mattrosno:
@alisonjoseph @andreancardona I uploaded the new libraries (backup) pictogram to our Sketch file in Box: https://ibm.ent.box.com/file/887827201496

image

jeanservaas:
@mattrosno talked to tracey... and to avoid confusion in tablet and mobile... let's go ahead and keep them for now and we'll test to see if it's a helpful label for people.

kingtraceyj:
@mattrosno jeannie brought this up and i missed this, but we are going to take out the library pictogram  to match the other black banners in the top section.

mattrosno:
@kingtraceyj thanks, updating now.

services/web-app/styles/vendor/_overrides.scss
```diff
@@ -5,13 +5,37 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+@use '@carbon/styles/scss/utilities/z-index' as z-index;
```
jdharvey-ibm, line 8: Josh pointed out to me that as of the most recent version of `@carbon/react`, we can pull this right from `@carbon/react/scss/utilities/z-index` to keep all of our sass uses coming exclusively from `@carbon/react`. I think we have this one and potentially one other usage of the zindex from styles in the web app that could be udpated.

mattrosno: I created a separate issue for this. Outside the scope of this PRs intent. https://github.com/carbon-design-system/carbon-platform/issues/431

jdharvey-ibm: 👍 We can toss this into next sprint.

alisonjoseph (approved):
🔥

jharvey10 (approved):

Issue #385, opened 2022-03-14
In-page navigation (sticky left)


mattrosno:
Reference pattern: https://www.ibm.com/standards/carbon/patterns/in-page-navigation

Use Dotcom's TOC: https://www.ibm.com/standards/carbon/react/?path=/story/components-table-of-contents--default

alisonjoseph:
Not sure we can get ` @carbon/ibmdotcom-react` to play nice <img width="818" alt="Screen Shot 2022-03-18 at 1 47 52 PM" src="https://user-images.githubusercontent.com/2753488/159065463-c74303ef-4664-4730-9059-35d2e0813ed1.png">

alisonjoseph:
After looking into using the dot com component we realized it wouldn't work for us due to v10/v11 compatibility issues, a long with the amount of layout overrides we would have to do because it is so opinionated on layout. 

@kingtraceyj is there a detailed spec for this, I didn't see it in the sketch file, but I could be missing it. Just want to be sure I have the color and spacing tokens correct, also happy to pair up quickly to tweak what I have built.

kingtraceyj:
Here is the spec: https://ibm.box.com/s/x72jtusuu91nab1usgmk4kbya1c2x2cm

![image](https://user-images.githubusercontent.com/20582179/159929562-f9aee3e7-2a85-4206-abae-f35170be78fb.png)

PR #401, opened 2022-03-18
refactor: grid update for page without sidenav

alisonjoseph:
Ref #385

In order to add the in page nav to the library pages we need to update the grid to span the full 16 columns on pages without a sidenav. 

#### Changelog

**Changed**

- Updates layout to not have an empty column when there is no sidenav
- Updates library detail and asset detail page grid to add left column for future content

#### Testing / reviewing

Make sure nothing changed with the grid on any pages.

### DIFF
--- a/services/web-app/layouts/layout/layout.js
+++ b/services/web-app/layouts/layout/layout.js
@@ -86,9 +86,9 @@ const Layout = ({ children }) => {
           </Theme>
           <Theme className={styles.body} theme="g10">
             <Grid as="main" className={styles.main} id="main-content">
-              <Column sm={4} md={8} lg={4}>
-                <Theme theme="white">
-                  {showSideNav && (
+              {showSideNav && (
+                <Column sm={4} md={8} lg={4}>
+                  <Theme theme="white">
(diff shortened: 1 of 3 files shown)


mattrosno (commented):
@alisonjoseph is this needed? The Carbon side nav is fixed position. To show the side nav on library pages, we can modify this line:

https://github.com/carbon-design-system/carbon-platform/blob/main/services/web-app/layouts/layout/layout.js#L46

alisonjoseph:
Yes, I think so. Currently pages without a side nav are only displaying in the 12 columns on the right, and we will need to put the in page nav in that left column. 

All this PR is doing is adding logic to display the content inside 16 columns if no sidenav and 12 if there is a sidenav. That way we can control the entire page grid for pages without the sidenav. 

We can pair up on Monday if this doesn't make sense.

francinelucca (dismissed):
This looks good to me, I don't perceive any UI difference and the grid looks good

mattrosno (approved):

francinelucca (approved):

PR #406, opened 2022-03-22
GitHub issue count 372

### LINKED ISSUE #372: Show open issue and PR counts per asset and link to them
**Old tasks**

- Add GitHub label to assets schema so we can query open issues and pull requests
- Query GitHub repo issues with that label when we fetch data and populate the data model with the results
  - Will this count get updated throughout the day with ISR
- Render issue and PR counts in asset detail page and link out to the list of GH issues or PRs so you can see those details
- Edge / empty states
- Documentation on how to use these labels

**New tasks**

- See comment

alisonjoseph:
Closes #372 

**Note**: Discussions are not currently part of the [GitHub search API](https://docs.github.com/en/rest/reference/search) this PR only adds issue counts. This is a future enhancement when we can look into using GitHubs GraphQL API. https://github.com/carbon-design-system/carbon-platform/issues/428

#### Testing / reviewing

Asset detail page should show correct number of open issues, and link to search by title in the issue repo. If there are no open issues it should show 0 and still link out.

Discussion card should just link to repo discussion search by title.

Screen Shot 2022-03-24 at 6 11 04 PM


services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js
```diff
@@ -175,56 +186,40 @@ const Asset = ({ libraryData }) => {
...
-                aspectRatio={{ md: '2x1', lg: '16x9', xlg: '2x1' }}
-                border={['sm', 'md', 'lg', 'xlg']}
-                href={libraryPath}
-              >
-                <dl>
-                  <dt className={dashboardStyles.label}>Coming soon...</dt>
-                  <dd className={dashboardStyles.labelLarge}>–</dd>
+                  <dt className={dashboardStyles.label}>Open issues</dt>
+                  <dd className={dashboardStyles.labelLarge}>
+                    {assetData.content.issueCount || 0}
```
Comment (francinelucca) [inline services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:198]:
```suggestion
                    {assetData.content.issueCount || '-'}
```

Comment (alisonjoseph) [reply services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:198]: We decided on the webex that 0 made more sense than just -, so a user can still click the link and see the issue count in github is actually 0.

Comment (francinelucca) [reply services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:198]: ohhh ok missed that 👍

mattrosno (changes requested):

andreancardona (dismissed):
yay!!

mattrosno (commented):
I'll review this afternoon - the changes from ISR to SSR for the asset detail page.

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js
```diff
@@ -92,6 +92,8 @@ const Asset = ({ libraryData }) => {
     }
   }
 
+  const githubRepoUrl = `http://${assetData.params.host}/${assetData.params.org}/${assetData.params.repo}`
+
```
Comment (jdharvey-ibm) [inline services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:96]:
```suggestion
  const githubRepoUrl = `https://${assetData.params.host}/${assetData.params.org}/${assetData.params.repo}`

```

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js
```diff
@@ -208,14 +201,14 @@ const Asset = ({ libraryData }) => {
...
+                href={`${githubRepoUrl}/discussions/?discussions_q=in%3Atitle+${assetData.content.id}`}
               >
                 <dl>
-                  <dt className={dashboardStyles.label}>Coming soon...</dt>
+                  <dt className={dashboardStyles.label}>Discussions</dt>
                   <dd className={dashboardStyles.labelLarge}>–</dd>
                 </dl>
                 <Svg32Github className={dashboardStyles.positionBottomLeft} />
-                {pathIsAbsolute('https://carbondesignsystem.com') && (
+                {pathIsAbsolute(githubRepoUrl) && (
```
Comment (jdharvey-ibm) [inline services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:211]: side note to consider renaming this function to `isPathAbsolute` as a more conventional boolean function name.

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js
```diff
@@ -231,7 +224,7 @@ Asset.propTypes = {
   libraryData: libraryPropTypes
 }
 
-export const getStaticProps = async ({ params }) => {
+export const getServerSideProps = async ({ params }) => {
```
Comment (mattrosno) [inline services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:227]: Long term, we'll want this to be ISR for optimal performance with on-demand regeneration. For for now, SSR seems totally fine for me. Our file system cache is caching GitHub requests, so each subsequent page view should read from the file system, get content, return server side rendered page. Each page request shouldn't fire off a GitHub request for this, so all good.

Cool with this @jdharvey-ibm?

If updating this page to SSR, lets do the same for `services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/index.js`

services/web-app/lib/github.js
```diff
@@ -310,10 +310,21 @@ const getLibraryAssets = async (params = {}) => {
...
+  const { host, org, repo } = asset.params
+  console.log(asset.prams, 'hi')
+  return (
+    (
+      await getResponse(host, 'GET /search/issues', {
+        q: `${asset.content.name}+repo:${org}/${repo}+is:issue+is:open+in:title`
+      })
+    ).total_count ?? 0
+  )
+}
```
mattrosno, line 326: - [ ] Remove console.log
- [ ] Add JSDoc comment
- [ ] Wrap getResponse in a try/catch like the other examples
- [ ] If there's a response, safely check for `total_count`
- [ ] Create new typedef for GitHubSearchResponse

I think something like this will work:

```suggestion
/**
 * Gets the GitHub open issue count for an asset using the asset's name searching only issue title
 * @param {import('../typedefs').Asset} asset
 * @returns {number}
 */
export const getAssetIssueCount = async (asset) => {
  const { host, org, repo } = asset.params

  /**
   * @type {import('../typedefs').GitHubSearchResponse}
   */
  let response = {}

  try {
    response = await getResponse(host, 'GET /search/issues', {
      q: `${asset.content.name}+repo:${org}/${repo}+is:issue+is:open+in:title`
    })
  } catch (err) {
    return null
  }

  return response?.total_count ?? 0
}
```

mattrosno (changes requested):

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js
Comment (jdharvey-ibm) [reply services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:227]: @mattrosno  Does nextjs provide a "deferred ISR" type of thing? As in: something that marks the pages as incrementally regenerated, but doesn't attempt to build them for the first time during the production build itself?

One of the problems we're hitting here is that at prod build time, we're slamming the github server with requests from the same personal access token to statically generate the initial copy of these pages. As the number of assets grows, it will become less and less feasible to do that and we're already basically at the limit of what a single token can do.

Also, once we have the data graph up and running, this data won't be available to the web-app at build time; it'll be maintained throughout the rest of the service infrastructure. When that happens, it won't be possible to obtain a comprehensive list of all components during the web app builds to populate `getStaticPaths`.

We can definitely keep a close eye on performance, but we'll likely have to use gSSP for most of these requests and rely on caching through the rest of the service infrastructure to keep things snappy.

Comment (mattrosno) [reply services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:227]: ISR doesn't build pages during the production build (`npm run build`.) Our issue is that the catalog pages are statically generated, and each of those catalog pages (and the libraries page) fetch all libraries (and their assets).

Comment (mattrosno) [reply services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:227]: Long term, we'll want library and detail pages ISR, and then catalog pages to be some hybrid where the catalog content is fetched either server side or client side when the page hydrates. So there's no dependency on GitHub data at production build time.

Comment (francinelucca) [reply services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:227]: now that I'm thinking about this, I think we could keep this as `getStaticProps` and get the same result; Matt's right these pages are not being built at build time. as long as we don't fetch the issueCount on`getLibraryData` we should be good.
TLDR: we can change `getServerSideProps` back to `getStaticProps` and re-add the `getStaticPaths` method

Comment (alisonjoseph) [reply services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:227]: What is the action item for this PR? 😄

francinelucca (dismissed):
🎉

services/web-app/lib/github.js
```diff
@@ -310,10 +310,33 @@ const getLibraryAssets = async (params = {}) => {
...
+/**
+ * Gets the GitHub open issue count for an asset using the asset's name searching only issue title
+ * @param {import('../typedefs').Asset} asset
+ * @returns {number}
+ */
+export const getAssetIssueCount = async (asset) => {
+  const { host, org, repo } = asset.params
+
+  /**
+   * @type {import('../typedefs').GitHubSearchResponse}
```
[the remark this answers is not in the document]
mattrosno: This helps state that we'll be depending on the `total_count` property. We aren't using `incomplete_results` right now, but might want to.

> To keep the Search API fast for everyone, we limit how long any individual query can run. For queries that [exceed the time limit](https://developer.github.com/changes/2014-04-07-understanding-search-results-and-potential-timeouts/), the API returns the matches that were already found prior to the timeout, and the response has the incomplete_results property set to true.

If incomplete result, we render an en-dash or something. Not needed, but that's there if we want it.

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js
Comment (mattrosno) [reply services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:227]: To clarify, are you and @jdharvey-ibm recommending the former `getStaticProps` in this page or the new `getServerSideProps`?

services/web-app/lib/github.js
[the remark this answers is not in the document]
alisonjoseph: Added 👍

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js
Comment (jdharvey-ibm) [reply services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:227]: Yeah I'm totally fine with adjusting any/all of the data fetching. As long as it's cached/reduces the number of requests that happen at build time, I'm good with whatever.

alisonjoseph:
Updated to use `getServerSideProps`

andreancardona (approved):

francinelucca (approved):

PR #407, opened 2022-03-22
fix: link style on asset detail page

alisonjoseph:
Bug noticed during dev sync

Screen Shot 2022-03-22 at 12 03 45 PM

#### Changelog

**Changed**

Updated class name on meta link large to use default Carbon link styles

#### Testing / reviewing

Check link style on asset detail page
Screen Shot 2022-03-22 at 12 02 38 PM

### DIFF
--- a/services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js
+++ b/services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js
@@ -111,7 +111,7 @@ const Asset = ({ libraryData }) => {
                   <dt className={dashboardStyles.label}>Library</dt>
                   <dd className={dashboardStyles.labelLarge}>{libraryData.content.name}</dd>
                 </dl>
-                <Link href={libraryPath} passHref={true}>
+                <Link href={libraryPath} passHref>
                   <CarbonLink className={clsx(dashboardStyles.metaLinkLarge)}>
                     {`v${libraryData.content.version}`}
                   </CarbonLink>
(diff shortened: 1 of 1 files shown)


mattrosno (approved):
Per docs, could go with just `passHref` instead of `passHref={true}`, but that works too!

https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-custom-component-that-wraps-an-a-tag

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js
```diff
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { Button, Column, Grid } from '@carbon/react'
+import { Button, Column, Grid, Link as CarbonLink  } from '@carbon/react'
```
Comment (mattrosno) [inline services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:8]: @alisonjoseph there's an extra space here, which is failing eslint for me. Let us know if you want to look at your VS Code settings so formatting happens on save.

mattrosno (changes requested):

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js
Comment (alisonjoseph) [reply services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:8]: Oops! I opened this PR via the GitHub api, that's why. 😉

mattrosno (dismissed):
🎉

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js
```diff
@@ -111,10 +111,10 @@ const Asset = ({ libraryData }) => {
                   <dt className={dashboardStyles.label}>Library</dt>
                   <dd className={dashboardStyles.labelLarge}>{libraryData.content.name}</dd>
                 </dl>
-                <Link href={libraryPath}>
-                  <a className={clsx(dashboardStyles.metaLink, dashboardStyles.metaLinkLarge)}>
+                <Link href={libraryPath} passHref>
+                  <CarbonLink className={clsx(dashboardStyles.metaLinkLarge)}>
```
Comment (mattrosno) [inline services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:115]:
```suggestion
                  <CarbonLink className={dashboardStyles.metaLinkLarge}>
```

Now that we are only using one class name here.

jharvey10 (approved):

francinelucca (approved):
LGTM

PR #410, opened 2022-03-22
fix: grid style fixes

### LINKED ISSUE #282: Figure out Grid narrow mode without Carbon v11 CSS grid not supporting narrow mode
### Browser

_No response_

### Description

Some of these components like the search bar, results bar, sort bar, pagination properly hang into the left gutter, but should not also hang into the right gutter like shown below.

We're overriding the grid condensed mode, and we need addition overrides to achieve hanging content per spec.

### Steps to reproduce

image

alisonjoseph:
Updating from Carbon v11 `0.15.2 `to `0.16.2` broke the grid in various places across the site.

This should also close #282 

#### Changelog

**New**

**Changed**

- 

**Removed**

- 

#### Testing / reviewing

Check grid alignment is correct everywhere.


mattrosno (dismissed):
Surprised I can be an approver since I pushed commits but hey, it's still looking good to me.

francinelucca (approved):
🎉

mattrosno (approved):

PR #422, opened 2022-03-23
fix(web-app): address array.map console errors

francinelucca:
Fixes:
image
image

### DIFF
--- a/services/web-app/layouts/layout/layout.js
+++ b/services/web-app/layouts/layout/layout.js
@@ -124,7 +124,7 @@ const Layout = ({ children }) => {
                           }
                           if (!data.path && data.items) {
                             return (
-                              <span key={i}>
+                              <div key={i}>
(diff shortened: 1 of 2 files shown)


services/web-app/layouts/layout/layout.js
```diff
@@ -124,7 +124,7 @@ const Layout = ({ children }) => {
                           }
                           if (!data.path && data.items) {
                             return (
-                              <>
+                              <span key={i}>
```
mattrosno, line 127: Can this be a `div` because some of the children like `h2` are display block?

mattrosno (changes requested):

services/web-app/layouts/layout/layout.js
francinelucca: Done!

alisonjoseph (approved):

mattrosno (approved):

PR #460, opened 2022-03-29
web-app: add no-use-before-define rule, fix env-vars file

### LINKED ISSUE #350: Monorepo: Make `'@typescript-eslint/no-use-before-define': 'error',` compliant

francinelucca:
Closes #350 

**Testing/Reviewing**

There are no code changes in this pr, other than those in `.eslintrc.json`, everything else I just moved around so functions would not be used before being defined, all logic stayed the same.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -25,6 +25,10 @@ module.exports = {
   },
   plugins: ['@typescript-eslint', 'jest', 'notice', 'simple-import-sort'],
   rules: {
+    '@typescript-eslint/no-use-before-define': [
+      'error',
+      { functions: false, classes: true, variables: false }
+    ],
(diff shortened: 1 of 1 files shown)


jdharvey-ibm:
I did some research into this linting rule to figure out if it's just enforcing things for the sake of enforcing them or if it's actually finding bugs and code smells. What I found was that in ES6, function declarations are hoisted to the top of modules and scope blocks, so the order of definition makes no difference. Classes are not hoisted, so this rule will enforce their definition before use in all cases. Variables aren't hoisted either, but the defaults of this rule were overly strict, even when the variables weren't in the same scope, so I turned that off as well. The variables rule will still catch use before definition within the _same_ scope, but no longer flag differing scopes.

jharvey10 (approved):

alisonjoseph (approved):

PR #464, opened 2022-03-30
feat: treeview library navigation

### LINKED ISSUE #452: Extend Carbon tree view with updated visuals to work in side nav

### LINKED ISSUE #453: Build side nav back, title, and links
- [x] Back
- [x] Title
- [x] Links (states)
- [x] Motion https://ibm.invisionapp.com/prototype/cl0jxmfjz0000jo01ub4ihcmw/play

alisonjoseph:
Closes #453 
Closes #452

## TODO 
- [x] update motion to push instead of slide-over
- [x] look into screen flash when you click back
- [x] refactor tree nav to recursively generate nodes vs. nesting

#### Changelog

**New**

- tree-nav component
  - style overrides for tree-view to match nav design spec
- Add back link to library detail page with css slide in/out
- Dummy nav data added 
- nav-main component (moved out of layout to reduce file complexity)
- nav-library component

**Changed**

- Update layout to display side nav on library detail page
  - tabindex is set to -1 on all active elements to effectively hide from screen reader when the secondary nav slides on top
- Updated layout context provider to manage nav state

**Removed**

- Remove breadcrumb from library detail page

#### Testing / reviewing

Navigate from Libraries page to a library detail page, and back. Test on mobile. Check a11y. Test with keyboard. 

Note: nothing is wired up to the treeview nav yet, this content will need to get added to the schema and pulled from the library repos. 

(animation/page transitions will be smoother/faster with production build vs. dev build)

h


francinelucca:
Visuals and desktop/mobile LGTM! I'm having a bit of trouble navigating with keyboard: it skips to the dashboard after the first option (can't seem to get to option 2) and it seems to be skipping through the main side nav options (assets, libraries, components, functions, patterns, templates, dataviz) which are not visually on the page before getting to the new side nav

alisonjoseph:
@francinelucca are you using the arrow keys? Treeview keyboard controls are different, https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2a.html

francinelucca:
> https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2a.html

whoops, thanks! arrows work for me 🎉 . I still have to "tab" through the unseen previous side nav options though before getting to this side nav which feels weird to me

alisonjoseph:
@francinelucca ack, yea that shouldn't be happening, that should be hidden. I'll take look.

services/web-app/layouts/layout/layout.js
```diff
@@ -34,14 +34,37 @@ export const LayoutContext = createContext()
 
 export const LayoutProvider = ({ children }) => {
   const [navData, setNavData] = useState([])
+  const [showSideNav, setShowSideNav] = useState(true)
+  const [librarySideNav, setLibrarySideNav] = useState(false)
+  const [isSideNavExpanded, toggleSideNavExpanded] = useState()
```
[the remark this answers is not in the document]
alisonjoseph: 👍

alisonjoseph:
@francinelucca updated, I had my logic reversed.

services/web-app/components/nav-tree/index.js
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright IBM Corp. 2021, 2022
```
mattrosno, line 2:
```suggestion
 * Copyright IBM Corp. 2022, 2022
```

mattrosno (commented):
When viewing a library, on mobile, when you open the navigation (from the hamburger icon), should the slide animation happen? I'm thinking no, as we're not entering the library page.

Related, I opened a new issue for side nav things I found. https://github.com/carbon-design-system/carbon-platform/issues/504

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/index.module.scss
```diff
@@ -32,3 +35,106 @@
...
+  position: relative;
+  background-color: theme.$background-selected;
+  color: theme.$text-primary;
+
+  &::after {
+    position: absolute;
+    top: 0;
+    left: 0;
+    display: block;
+    width: 4px;
```
Comment (andreancardona) [inline services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/index.module.scss:94]:
```suggestion
    width: spacing.$spacing-02;
```

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/index.module.scss
```diff
@@ -32,3 +35,106 @@
...
+
+  &:focus {
+    outline: 2px solid theme.$focus;
+    outline-offset: -2px;
+  }
+}
+
+.library-nav-in {
+  animation: slide-in motion.$duration-moderate-01 motion.motion(entrance, productive) forwards;
+  transform: translateX(-256px);
```
Comment (andreancardona) [inline services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/index.module.scss:122]: Do we want to make this a token since we are using it 2x 🤷🏽‍♀️ (line 138)

andreancardona (dismissed):
Approving since my comments are pretty insignificant and the mobile bugs are already documented :)

Tested and looks greatt!! 🎉

alisonjoseph:
@jeanservaas oh yea, right now the treeview doesn't actually link anywhere so you're able to select it. If I add links that won't happen.

alisonjoseph:
@mattrosno updated so the slide in/out only happens at desktop

kingtraceyj:
@alisonjoseph jeannie and i just went through some motion specs and i think we might want to change the orientation to be something more like this: http://dev.maggieryder.com/storybook/?selectedKind=JQuery%20dl-menu%20copy&selectedStory=sliding%20transition&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

i can jump on a screenshare or something to show what we are thinking.

alisonjoseph:
@kingtraceyj I think that makes sense, so instead of just sliding over, the library nav slides in and pushes the main nav offscreen? This looks slower than our motion specs, should we keep the speed at `$duration-moderate-01` (150ms).

kingtraceyj:
@alisonjoseph yeah. i think we want to keep our speeds the same. theirs is super slow. 

the background would stay the same and the nav items move left and right off screen.

services/web-app/components/nav-tree/nav-tree.js
```diff
@@ -0,0 +1,76 @@
...
+import PropTypes from 'prop-types'
+
+import styles from './nav-tree.module.scss'
+
+const NavTree = ({ activeItem, items = [], label }) => {
+  const router = useRouter()
+  const toggle = function () {
+    return null
+  }
+
```
jdharvey-ibm, line 18: An option for reducing the duplication here would be to recursively generate the hierarchy of TreeNodes. That would enable the max depth to be configurable as an added bonus :)

services/web-app/components/nav-tree/nav-tree.module.scss
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright IBM Corp. 2021, 2022
```
jdharvey-ibm, line 2:
```suggestion
 * Copyright IBM Corp. 2022, 2022
```

services/web-app/components/nav-tree/nav-tree.module.scss
```diff
@@ -0,0 +1,84 @@
...
+
+  :global(.cds--tree-parent-node__toggle::after) {
+    position: absolute;
+    top: spacing.$spacing-03;
+    right: spacing.$spacing-05;
+    display: block;
+    width: spacing.$spacing-05;
+    height: spacing.$spacing-05;
+    // chevron icon
+    background-image: url("data:image/svg+xml,%3Csvg focusable='false' preserveAspectRatio='xMidYMid meet' xmlns='http://www.w3.org/2000/svg' fill='currentColor' width='16' height='16' viewBox='0 0 16 16' aria-hidden='true' class='cds--tree-node__icon'%3E%3Cpath d='M8 11L3 6 3.7 5.3 8 9.6 12.3 5.3 13 6z'%3E%3C/path%3E%3C/svg%3E%0A");
```
jdharvey-ibm, line 46: Is this something we could break out into a React component instead?

services/web-app/layouts/layout/layout.js
```diff
@@ -56,12 +88,27 @@ const Layout = ({ children }) => {
...
+    setLibraryNavSlideOut(false)
+  }, [setShowSideNav, setLibrarySideNav, setLibraryNavSlideOut, router.pathname])
+
+  const onClickSideNavExpand = () => {
+    toggleSideNavExpanded(!isSideNavExpanded)
+  }
+
+  const backLink = () => {
+    setLibraryNavSlideOut(true)
+    setTimeout(() => router.push('/assets/libraries'), 150)
```
jdharvey-ibm, line 104: minor: magic number could be defined as a constant at the top describing its purpose.

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/index.module.scss
```diff
@@ -7,6 +7,8 @@
 
 @use '@carbon/react/scss/breakpoint' as breakpoint;
 @use '@carbon/react/scss/spacing' as spacing;
+@use '@carbon/react/scss/theme' as theme;
+@use '@carbon/react/scss/type' as type;
```
Comment (jdharvey-ibm) [inline services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/index.module.scss:11]: minor: Remove these unused imports

Comment (alisonjoseph) [reply services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/index.module.scss:11]: ah good catch, leftover from me moving things around

services/web-app/components/nav-tree/nav-tree.module.scss
alisonjoseph: Agree this is a little funky, what we really need is to extend the carbon treeview so we can apply a custom icon for the toggle, but until we do that this was the interim solution I came up with. I can see if I can add the icon as a child of TreeNode if the node has children. 🤔

kingtraceyj:
this is looking good @alisonjoseph! The main overview page is no longer selected by default once a user gets into a library.

and then, when back to all libraries is clicked, it's little laggy to get back to the libraries list page. is there a delay added or is just that loading time?

alisonjoseph:
@kingtraceyj the lag is because its a dev environment, however there is a weird flash I'm looking into now.

services/web-app/components/nav-tree/nav-tree.js
```diff
@@ -11,11 +11,31 @@ import PropTypes from 'prop-types'
...
+        key={nodeProps.title}
+        onToggle={toggle}
+        onClick={() => nodeProps.path && router.push(nodeProps.path)}
+        {...nodeProps}
+      >
+        {renderTree({ nodes: items, label, value, id, key, onClick, onToggle })}
+      </TreeNode>
+    ))
+  }
+
```
jdharvey-ibm, line 38: 😎 🔥

francinelucca: Nice! 🎉

francinelucca (dismissed):

alisonjoseph:
@kingtraceyj fixed the flash, and made a screen recording of a prod build showing no delay 
https://user-images.githubusercontent.com/2753488/163278746-527974a8-80a1-4789-b279-8d423c59d9e7.mov

kingtraceyj (dismissed):
looks good to go! 🚀

francinelucca (approved):
still LGTM

kingtraceyj (approved):

PR #509, opened 2022-04-06
Flucca/feat/library assets

### LINKED ISSUE #375: Library assets page
**Design**

- [x] Red lines https://ibm.ent.box.com/file/913340274022

**Dev**

- [x] Sort
- [x] Assets table
  - [x] Color coded type tags
  - [x] Hover states
  - [ ] Empty states
  - [ ] Link out per row

Notes:

1. No pagination needed at the moment. We can add pagination later if needed by our data fetching service.

francinelucca:
Closes #375

No side nav yet, need #464 to be merged in 
Design: https://ibm.ent.box.com/file/913340274022

#### Testing / reviewing

Example url: http://localhost:3000/assets/carbon-charts/library-assets
Test all breakpoints against design, make sure linkout works on asset rows, test empty state

https://user-images.githubusercontent.com/40550942/162341860-e04c5109-e9d0-4cca-97c2-29effe26ddde.mov

https://user-images.githubusercontent.com/40550942/162341865-46765bf2-8c93-4df5-b02e-e66d822530b8.mov


francinelucca:
Ready for first review.
Missing: clickable row (entire row). needs further discussion, might want to split this into a separate ticket.

francinelucca:
@kingtraceyj added some screen records so you can take a look, happy to jump on a huddle/call as well and review!

alisonjoseph:
@francinelucca quickly looking at this made me realize some updates I need to make to #375 to actually wire it up as a working nav

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/assets/index.js
```diff
@@ -0,0 +1,233 @@
+/*
+ * Copyright IBM Corp. 2021, 2022
```
Comment (alisonjoseph) [inline services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/assets/index.js:2]:
```suggestion
 * Copyright IBM Corp. 2022, 2022
```

kingtraceyj (changes requested):
looing good francine! there are just a few things i want to double check.

**Arrow color:**
I just want to double check that arrow color is `$interactive` i forgot that token in the redlines

This was also in another redline and i did not get it into the library redline but we put 24px padding on xlg breakpoints and 16px at lg, md, and sm break points within the text container. i just wanted to double check to see if that was happening here. 
image

This truncated in the Tags column gets lost on tablet and mobile. It's probably best to not truncate and start scroll early. We could look into adding a hover tooltip but that might not be accessible if a user cannot tab into a table to get that information.
image

also small, it looks like there is a tiny gap between the sort and the table and there should be no space: 
image

francinelucca:
@kingtraceyj addressed comments + added text padding, ready for second review
image

kingtraceyj (changes requested):
hey francine! the changes are looking good. one thing, now that we see it in code, is that the assets title is a little redundant with the nav in the header. we are going to change the title of the page from "Assets" to "Library assets" and take out the "All library assets" title above the table. 

I also wanted to check the width of the dropdown, it should be 4 columns and hang 16px. I can't get my carbon dev tools extension to work so i can't see the grid but it should work like this:
image

francinelucca:
@kingtraceyj updated lg columns! hang lgtm

kingtraceyj (dismissed):
looks good to go!

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/assets/index.js
```diff
@@ -0,0 +1,234 @@
...
+    <>
+      <NextSeo {...seo} />
+      <Grid className={styles.libraryAssetsContainer}>
+        <Column sm={4} md={8} lg={{ start: 5, span: 12 }}>
+          <PageHeader title={seo.title} />
+        </Column>
+        <Column sm={4} md={8} lg={4}>
+          {/* In page nav  */}
+        </Column>
+        <Column sm={4} md={8} lg={12}>
```
Comment (alisonjoseph) [inline services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/assets/index.js:133]:
```suggestion
        <Column sm={4} md={8} lg={{ start: 5, span: 12 }}>
```

I think we can just do this here instead, the library side nav will be controlled from the layout.

alisonjoseph (commented):
This looks good to me, other than a super minor comment!

kingtraceyj (dismissed):
looks good!

services/web-app/components/catalog-item/catalog-item-meta/catalog-item-meta.js
```diff
@@ -0,0 +1,60 @@
...
+
+import StatusIcon from '@/components/status-icon'
+import { status } from '@/data/status'
+import { getLicense } from '@/utils/schema'
+
+import styles from './catalog-item-meta.module.scss'
+
+const CatalogItemMeta = ({ asset, className, properties }) => {
+  const renderStatus = () => {
+    const statusKey = asset?.content.status?.key ?? asset?.content.status ?? 'draft'
```
jdharvey-ibm, line 20: It seems like this implies that the status key can either come in on the `status` field or on the `status.key` field. Is there any way to adjust that so the key is always present on the same field in the asset?

services/web-app/next.config.js
```diff
@@ -80,6 +80,16 @@ module.exports = withMDX({
...
+      rewrites.push({
+        source: `/assets/${slug}/assets`,
+        destination: `/assets/${library.host}/${library.org}/${library.repo}/${slug}/latest/assets`
+      })
+
+      rewrites.push({
+        source: `/assets/${slug}/:ref*/assets`,
+        destination: `/assets/${library.host}/${library.org}/${library.repo}/${slug}/:ref*/assets`
+      })
+
```
jdharvey-ibm, line 92: Am I right that this would mean we'd have URLs like:
`next.carbondesignsystem.com/assets/some-thing/assets`?

If so, could we maybe chat about that to see if we're heading down a weird path?

francinelucca: yeah so, a library's url is "/assets/library-slug" and this is a page for a library's assets, so I put it under "/assets/library-slug/assets" , I see how that's weird but it also makes sense to me to put the library's "assets" page under [library-url]/assets 🤔

services/web-app/components/catalog-item/catalog-item-meta/catalog-item-meta.js
francinelucca: yeah so this is how we have it defined in our schema right now (see comment on: https://carbon-design-system.github.io/carbon-platform/resource-schemas.html#asset-status). We can open up a ticket to standardize at the `getLibraryAssets()` level if desired but that'd fall out of the scope of this PR; I'm just moving this out of `services/web-app/components/catalog-item/catalog-item.js` and into it's own component so I can re-use it

services/web-app/next.config.js
alisonjoseph: The url felt a little weird to me too when I was testing, but agree that unless naming of something changes, it makes sense. Tagging @aubrey-oneal and @kingtraceyj just as an FYI here.

services/web-app/components/catalog-item/catalog-item-meta/catalog-item-meta.module.scss
```diff
@@ -0,0 +1,34 @@
...
+  display: inline;
+
+  &:not(:first-child) {
+    margin-left: spacing.$spacing-03;
+  }
+}
+
+.meta-icon {
+  position: relative;
+  top: 3px;
```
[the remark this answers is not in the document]
alisonjoseph: The svg is just a circle, so not sure that editing the svg would make sense? I think adding a comment that this is to visually align with the text is fine here.

services/web-app/next.config.js
mattrosno: I'll create a separate issue for side nav (and URL) naming to not block this PR.

kingtraceyj: @mattrosno We are trying to create a distinction between the UI Shell 'Assets' and the library assets. I think it's OK to get specific with what assets you are viewing. You aren't looking at all assets just the specific library ones. Using this logic we could go down the path of putting the word library in front of each nav item, but there's no real need since it's unrelated to the top nav item. We've had multiple people, seperately comment on that it's a little confusing to have both the top level nav item say assets as well as the side nav say assets. i think getting specific on this one category adds clarity.

alisonjoseph: Updating the first item to be "Library assets" and the url to` /assets/carbon-charts/library-assets` is simple enough, and adds clarity.

mattrosno: I'm okay with that for this PR. We're looking at side nav in a separate issue to ensure all of the other items, and their order, are right.

francinelucca: fixed! new sample url: http://localhost:3000/assets/carbon-charts/library-assets

services/web-app/components/catalog-item/catalog-item-meta/catalog-item-meta.js
francinelucca: Summarized in #562

alisonjoseph (dismissed):

alisonjoseph (approved):

services/web-app/next.config.js
jdharvey-ibm: Some notes to toss in the URL vs nav issue:

Can you refresh my memory about the root-most `/assets`? Are we saying that a Library is an Asset and also Libraries can *have* Assets, some of which might be other Libraries? That's what the current URL scheme implies, but I want to confirm that's the intent.

Is `/assets/button` ever a valid url? Or is it always `/assets/some-library-name/button`?

If it's only ever `/assets/some-library`, then I'd vote for being more explicit in the URL about the type of object you're viewing: `/libraries/carbon-charts`. Or even more explicitly: `/assets/libraries/carbon-charts/assets`.

At the moment, there's a mismatch between the breadcrumb and the URL (see screenshot). Assets, Libraries, and a specific library are referenced as navigation contexts on the page, but the URL is missing the indication of the specific library being nested under the set of all libraries.

If a library is not considered as an asset itself and is only a container of *other* assets, then I'd vote for `/libraries/carbon-charts/assets` as opposed to `/assets/libraries/carbon-charts/assets` so that the left nav becomes the more definitive source for the URL format.

(Marking as unresolved so the thread doesn't get lost until we copy/paste into an issue).

image

services/web-app/components/catalog-item/catalog-item-meta/catalog-item-meta.module.scss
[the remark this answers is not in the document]
jdharvey-ibm: Some other food for thought: adjusting the vertical alignment to something like `vertical-align: middle;` should get this to the right spot as well.

services/web-app/components/catalog-item/catalog-item-meta/catalog-item-meta.js
jdharvey-ibm: Oof, that definitely makes parsing the schema harder, but I see the value in the convenience for the end user to be able to forego the note if all they want to add is the key. Thanks for opening the issue!

jharvey10 (approved):

PR #658, opened 2022-04-27
fix: storybook layout for next pages

mattrosno:
We have a custom layout global control to set full screen or padded. We don't use the out-of-box Storybook layout options because we're setting a global theme (e.g. g10) and we need the padding to be inside that, not around it.

image

This PR:

- Adds comments to the full screen / padded custom global styling
- Ensures the "page" stories always renders full screen because if the preview is padded, it pushes page content in and will lead to falsely-reported bugs
- Moves "page" story config and layout wrapper components to a new `layout.js` file

### DIFF
--- a/services/web-app/.storybook/layout.js
+++ b/services/web-app/.storybook/layout.js
@@ -1,0 +1,41 @@
+/*
+ * Copyright IBM Corp. 2022, 2022
+ *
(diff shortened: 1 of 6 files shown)


francinelucca (approved):
lgtm

alisonjoseph (approved):

Issue #643, opened 2022-04-26
Tabs in mobile are being cut off


andreancardona:
### Description

Tabs in mobile are being cut off
<img width="308" alt="Screen Shot 2022-04-26 at 1 56 46 PM" src="https://user-images.githubusercontent.com/32720851/165373826-e9512714-81b2-4016-8443-5944abcfe1c8.png">

### Steps to reproduce

Navigate to: https://next.carbondesignsystem.com/assets/carbon-react/latest/accordion

Resize for mobile screen size

PR #671, opened 2022-04-29
fix: tabs sticky and mobile bug

### LINKED ISSUE #505: Page tabs should be sticky when scrolling
### To do

-[ ] Verify scroll padding (the CSS variable) used for anchor link scroll-tos

(We also have an example of a JS scroll event for the catalog sticky filter if we need do to anything in JS too, not just CSS.)

### Browser

_No response_

### Description

image

The page tabs (Usage, Design, etc.) should be sticky, so when scrolled, they are always visible like:

image

### Steps to reproduce

See above.

andreancardona:
Closes #505   &  #643 

- This PR makes sure tabs are sticky 
- It also makes sure the sudo element expands full vw at all. breakpoints

NOTE: thank you @francinelucca for helping me on this one!!  💯 

## TESTING:

- [ ] make sure sticky tabs stay sticky
- [ ] make sure there are no visual regressions with tabs at any breakpoint


services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].module.scss
```diff
@@ -43,3 +44,28 @@
     width: 50%;
   }
 }
+
+.tabs-column {
+  position: sticky;
+  z-index: 9000;
```
Comment (francinelucca) [inline services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].module.scss:50]: you can import like: `@use '@carbon/react/scss/utilities/z-index' as z-index;`
```suggestion
  z-index: z-index.z('overlay');;
```

francinelucca (changes requested):
minor CRs.

it doesn't look like our calculations are working for the line span on max breakpoint 😢 , let's sync on monday 
image

francinelucca (approved):
💃

alisonjoseph (approved):
🔥

PR #685, opened 2022-05-04
fix(web-app): mdx styling bugs

alisonjoseph:
Closes #

Testing out some real mdx content and noticed some spacing/layout bugs. 

- brings over the default background color we have on images in mdx from the Carbon website.
- Fix for anchorlinks not splitting into columns at correct count.
- fix for videos not having correct spacing, margin, wraps inside a grid component
- fix resource card border replaced with box shadow to fill in 1px condensed grid space

#### Testing / reviewing

Hard to test most of these without a full page from Carbon, make sure nothing in the MDX sandbox seems broken. http://localhost:3000/mdx


services/web-app/components/anchor-links/anchor-links.js
```diff
@@ -21,7 +21,7 @@ import styles from './anchor-links.module.scss'
  * content is long enough.
  */
 const AnchorLinks = ({ children, small, className }) => {
-  const isColumn = Children.count(children) > 6
+  const isColumn = Children.count(children) > 9
```
francinelucca, line 24: after how many links is this supp to break into a new column? it's breaking at 4 for me, correct?
image

services/web-app/components/markdown/image.module.scss
```diff
@@ -21,3 +25,8 @@
     object-fit: contain;
   }
 }
+
+// targets images on the page not nested inside other components
+:global([class*='layout_column-content']) > .image-container {
```
francinelucca, line 30: is this what it's supp to look like at small breakpoint? margin left and then take entire width after? feels a bit weird 🤔 
image

services/web-app/components/anchor-links/anchor-links.js
alisonjoseph: This is correct

services/web-app/components/markdown/image.module.scss
alisonjoseph: Nope, that's weird. I'll fix

alisonjoseph: Updated!

alisonjoseph:
Going to include this in my mdx content migration PR so its easier to test

PR #700, opened 2022-05-06
feat(web-app): medium posts

### LINKED ISSUE #540: MediumPosts MDX component

francinelucca:
Closes #540 

TODO

- [x] Images from medium not loading
- [x] can I cache the response
- [ ] use graphQl endpoint

#### Testing / reviewing

http://localhost:3000/mdx/medium-posts
http://localhost:6006/?path=/docs/mdx-mediumposts--default
Compare with:
https://carbondesignsystem.com/#latest-news-and-articles


francinelucca:
Having problems loading thumbnail from medium 😭 , will sync with Joe tomorrow

francinelucca:
Won't merge this until we get graphQl endpoint to retrieve posts (coming soon 💃), but'd still like to get some eyes/reviews on this 🙏

services/web-app/components/article-card/article-card.module.scss
```diff
@@ -62,6 +62,10 @@
   @include type.type-style('label-01');
```
alisonjoseph, line 62:
```suggestion
  @include type.type-style('label-01');
```
```suggestion
```

We should be able to remove this, since we're setting it on the paragraph.

services/web-app/components/medium-posts/medium-posts.module.scss
```diff
@@ -0,0 +1,42 @@
...
+    letter-spacing: 0;
+    line-height: 1rem;
+  }
+  @include breakpoint.breakpoint('xlg') {
+    font-size: 1rem;
+    font-weight: 400;
+    letter-spacing: 0;
+    line-height: 1.375rem;
+  }
+}
```
alisonjoseph, line 34:
```suggestion
```

Do we need these lines? I removed them locally and didn't see any difference? What am I missing? I think we should be good to just pull styles directly from Article Card.

alisonjoseph, line 34:
```suggestion
```

Do we need these lines? I removed them locally and didn't see any difference? What am I missing? I think we should be good to just pull styles directly from Article Card.

alisonjoseph (commented):
This is looking good!

services/web-app/components/medium-posts/medium-posts.js
```diff
@@ -0,0 +1,128 @@
...
+
+const MediumPosts = ({ postLimit = 3, cardProps, dummy, ...rest }) => {
+  const isCancelled = useRef(false)
+  const [allPosts, setAllPosts] = useState(dummy ? dummyArticles : [])
+
+  const isMd = useMatchMedia(mediaQueries.md)
+
+  useEffect(() => {
+    async function getMediumArticles() {
+      const articlesRes = await fetch('http://localhost:3000/api/get-medium-feed') // TODO: replace this with graphql endpoint call
```
jdharvey-ibm, line 52: Once we have this on the data graph, we should call it in gSSP or gSP with a revalidation interval :)

services/web-app/lib/file-cache.js
```diff
@@ -96,6 +96,19 @@ export const getResponse = (host, route, options) => {
...
+ * calls supplied callback to retrieve info
+ * @param {string} key - Key to use to save/retrieve cached response
+ * @param {Function} callback - Function to call to get info in case of invalid cache
+ * @returns {Promise<object>} GitHub API response data
+ */
+export const cacheResponse = (key, callback) => {
+  return diskCache.wrap(key, () => {
+    return callback()
+  })
+}
```
jdharvey-ibm, line 110: Would this work like this?

```suggestion
export const cacheResponse = (key, callback) => {
  return diskCache.wrap(key, callback)
}
```

services/web-app/package.json
```diff
@@ -68,6 +69,7 @@
     "remark-gfm": "^3.0.1",
     "remark-unwrap-images": "^3.0.1",
     "rimraf": "^3.0.2",
+    "rss-parser": "^3.12.0",
```
jdharvey-ibm, line 72: TODO for us to remove these packages once this is in the data graph

services/web-app/components/medium-posts/medium-posts.js
francinelucca: this is a component , not a page so it doesn't take gssp 😕

jdharvey-ibm: Oh haha. Good point. Then I guess passing it in as a prop from the containing page?

services/web-app/components/medium-posts/medium-posts.module.scss
francinelucca: LGTM too , thanks!!

services/web-app/components/medium-posts/medium-posts.js
francinelucca: yeah, depends if we'll consume it from a local mdx file or remote. if we do local I think we can add a gssp to it and consume the data then pass as prop

jdharvey-ibm: Thinking about this some more based on the current state of our repo and mdx stuff...

In a similar way to how we'd talked about discovering images in MDX, optimizing them, and storing them as source sets "on the side" for later, this seems like a similar type of use case: MDX is ingested, certain nodes in the AST require additional processing before display, processing happens, AST nodes are updated accordingly based on the results, updated MDX (or JSX) is passed on through to the web-app from which it was originally requested.

This might put us back into the realm of needing an actual MDX service... Something that could be asked for "ready-to-go" MDX, and behind the scenes could be parsing it, calling out to other services to resolve/generate data, and compiling the results before sending back to the web-app.

Relative to this PR, discovering a `MediumPosts` component during MDX processing could trigger a query on the data-graph to get the post data (i.e. the props needed to render the list of medium articles).

Then we'd benefit from any caching done by the medium service to speed things up. 🚀 

The overall benefit of doing it like that as opposed to via fetch request in an effect is that it could be server-side rendered, which would be good for cross-linking and SEO. (and not needing skeletons while the content is loading).

francinelucca: If it were up to me i'd go the medium-service -> gssp route just for the sake of simplicity, is there other advantages to doing the mdx-processing service instead?

francinelucca:
handled in #1445

Issue #690, opened 2022-05-05
Fetch MDX file content from a remote repository


mattrosno:
As a proof of concept, let's have a short-lived test page that as a specified remote MDX file path, and renders that MDX content.

E.g.: https://github.com/carbon-design-system/carbon-website/blob/main/src/pages/components/accordion/style.mdx

This test page should not have any imports or unsupported MDX components so it acts as a good baseline.

mattrosno:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-Development-624c3fc8b7972e001130fef7/issues/carbon-design-system/carbon-platform/690?planning-poker) with ZenHub @alisonjoseph @andreancardona @francinelucca @jdharvey-ibm

mattrosno:
Closed via https://github.com/carbon-design-system/carbon-platform/pull/711.

Issue #693, opened 2022-05-05
Optimize images used in remote markdown


mattrosno:
When rendering remote MDX content:

1. Find all images in the content
2. Construct absolute URLs to each image
3. Run each image through the plaiceholder library to generate placeholder base 64-encoded images and get image dimensions

THEN

1. Pass the placeholder image data into the page
1. When the page renders, do a lookup for each image and use the placeholder data where there's a match

OR

1.  Modify at the AST level so the returned markdown uses Next.js `Image` component with the plaiceholder props merged in

So the Next.js `Image` component can optimize the image because there are knowns heights and widths and blur datas.

See: https://github.com/carbon-design-system/carbon-platform/pull/551

And: https://github.com/carbon-design-system/carbon-platform/issues/547

Implement remote images option B from that issue/PR using plaiceholder.

The good stuff: https://github.com/carbon-design-system/carbon-platform/issues/547#issuecomment-1097297125

mattrosno:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-Development-624c3fc8b7972e001130fef7/issues/carbon-design-system/carbon-platform/693?planning-poker) with ZenHub @alisonjoseph @andreancardona @francinelucca @jdharvey-ibm

mattrosno:
What safeguards do we need to put in place to prevent malicious behavior? (E.g. verify that all images are actually coming from that library's repo? Etc.)

francinelucca:
Closing, new media mapper issues defined in https://github.com/carbon-design-system/carbon-platform/issues/989 after technical design

PR #711, opened 2022-05-10
feat(web-app): render remote mdx

### LINKED ISSUE #321: Remote MDX functionality
A key aspect of the platform's content mesh is fetching markdown (MDX files) from remote GitHub repositories and rendering in the web app. Assuming we use https://github.com/hashicorp/next-mdx-remote, there are import / export caveats to consider: https://github.com/hashicorp/next-mdx-remote#import--export

This makes a lot of sense. When pulling in remote MDX content not coming from the carbon-platform repo, we can't control that content, and can't let MDX import and run arbitrary components. Also, when the MDX content is remote, it doesn't run through Webpack.

There are multiple remote mdx instances that can break the mdx rendering in-app as well, we need a process to handle those.

We'll need a scrubbing process so when we fetch MDX file content through GitHub's API, we clean it to remove/modify unwanted or unrenderable components.

There is also some media handling that needs to be done in order to safely and accurately load media such as images and videos.

So the platform can safely render it.

francinelucca:
Closes #321 

#### New

- getRemoteMdxData function 
- RemoteMdxLoader component
- *temporary* github unoptimized image resolver <- this will be adressed in #693 
- Sample MDX page rendering from remote repo (hardcoded params)

#### Testing / reviewing

http://localhost:3000/samples/remote-mdx-test
https://carbondesignsystem.com/components/accordion/style


alisonjoseph (commented):
This is working great!

Question, if we're using `next-mdx-remote` are we able to use that for local mdx files also instead of `@next/mdx`. I thought I read it was an either/or thing, could be out of scope for this PR, but not sure if we should look into that?

francinelucca:
> This is working great!
> 
> Question, if we're using `next-mdx-remote` are we able to use that for local mdx files also instead of `@next/mdx`. I thought I read it was an either/or thing, could be out of scope for this PR, but not sure if we should look into that?

@next/mdx doesn't support remote loading, we *could* use next-mdx-remote for our local pages but then we'd have to create a .js page for each and use a getStaticProps method to import the mdx file and serialize the contents 🤔 (https://github.com/hashicorp/next-mdx-remote#examples) , not sure we'd want to do that?

alisonjoseph (approved):
> we could use next-mdx-remote for our local pages but then we'd have to create a .js page for each and use a getStaticProps method to import the mdx file and serialize the contents

@francinelucca hmm yea, if we'd have to do that for each page that doesn't seem like it would work.

andreancardona (approved):
🔥

mattrosno:
This should have closed https://github.com/carbon-design-system/carbon-platform/issues/690 not the epic. Fixing now. NBD.

PR #739, opened 2022-05-17
feat: remote mdx - optimize images

### LINKED ISSUE #693: Optimize images used in remote markdown

francinelucca:
Closes #693 

#### Testing / reviewing

Should use next/image and be optimized (not have image_container class)
http://localhost:3000/samples/remote-mdx-test


mattrosno (commented):
I didn't test this, but it looks awesome. Correct if we set height and width and a blur data URL we don't need our custom class names on the image.

alisonjoseph (dismissed):
Tested and images look good to me! I did notice something with image sizes compared to how they display on the Carbon website and created https://github.com/carbon-design-system/carbon-platform/issues/742

services/web-app/utils/mdx-image-optimizer-plugin.mjs
```diff
@@ -0,0 +1,42 @@
...
+import { getPlaiceholder } from 'plaiceholder'
+import { visit } from 'unist-util-visit'
+
+const mdxImageOptimizerPlugin = (dirPath) => async (tree) => {
+  const matches = []
+  visit(tree, { type: 'image' }, (node) => matches.push(node))
+
+  const promises = matches.map(async (node) => {
+    // construct full image path (with repo url)
+    const fullPath = `${dirPath}/${node.url}?raw=true`
```
jdharvey-ibm, line 16: Would this break if you encounter an image that already has a fully qualified URL as its source?

Also, does the allow you to do something like `../../../../my-different-repo/some-vulnerable-image-i-want-to-display-in-everyones-browser.png` by specifying a maliciously crafted URL in your markdown?

Better question: Can we enforce that the source markdown file and the target org/repo combo is the same when we've determined that the image is coming from github?

services/web-app/utils/mdx-image-optimizer-plugin.mjs
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright IBM Corp. 2022, 2022
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import { getPlaiceholder } from 'plaiceholder'
+import { visit } from 'unist-util-visit'
+
+const mdxImageOptimizerPlugin = (dirPath) => async (tree) => {
```
jdharvey-ibm, line 10: It might be good to document the assumptions and the contract of this plugin. Based on the presence of `raw=true` below, it seems like the intent might be to have this only work with github images, so if that's the case I think it's important to document all of that and make the contract of this function crystal clear.

francinelucca: Added comments on top of declaration !

services/web-app/utils/mdx-image-optimizer-plugin.mjs
francinelucca: > added check to not alter url if image has a fully qualified url
> added check to validate repo hasn't changed when constructing absolute url from relative source

package-lock.json
```diff
@@ -39919,6 +39920,33 @@
...
+        "unist-util-is": "^5.0.0"
+      },
+      "funding": {
+        "type": "opencollective",
+        "url": "https://opencollective.com/unified"
+      }
     }
   }
-}
+}
```
jdharvey-ibm, line 39952: this is kind of suspicious that there's no newline at the end. Could you run an `npm install` on this branch see if it re-adds that?

services/web-app/utils/mdx-image-optimizer-plugin.mjs
```diff
@@ -4,35 +4,69 @@
...
+        .slice(0, 4)
+    ) ===
+    JSON.stringify(
+      url2
+        .split('/')
+        .filter((chunk) => !!chunk)
+        .slice(0, 4)
+    )
+  )
+}
```
jdharvey-ibm, line 29: We should chat through this. Is there an opportunity to use the URL lib here or a regular expression instead?

services/web-app/utils/mdx-image-optimizer-plugin.mjs
jdharvey-ibm: Can we convert them to jsdoc? :) 
That might entail breaking this up into separate functions. Right now it's a function that returns a function that modifies the AST. That might be hard to jsdoc as-is.

services/web-app/utils/mdx-image-optimizer-plugin.mjs
jdharvey-ibm: Would this be vulnerable to a URL such as `https://github.com/carbon-design-system/carbon/packages/react/../../../malicious-org/bad-repo/uh-oh.jpg`?

jharvey10 (changes requested):

services/web-app/utils/mdx-image-optimizer-plugin.mjs
francinelucca: sure! I don't see how we could use the url lib but I'm open to try a regex I just don't know how...
 the reason I'm having to do this is because the dirPath looks something like:
`https://github.com/carbon-design-system/carbon-website/blob/main/src/pages/components/accordion` and after I create the imagePath with path.join it looks like this `https:/github.com/carbon-design-system/carbon-website/blob/main/src/pages/components/accordion/images/accordion-style-2.png/?raw=true` so I need to compare the first four chunks of the url and I can't do that directly because for some reason path.join is stripping the second '/' after https

I am using path.join now, so a url like `https://github.com/carbon-design-system/carbon/packages/react/../../../malicious-org/bad-repo/uh-oh.jpg` will be converted to `https://github.com/carbon-design-system/malicious-org/bad-repo/uh-oh.jpg` and that will not pass the check

package-lock.json
francinelucca: I did!

services/web-app/utils/mdx-image-optimizer-plugin.mjs
francinelucca: Done!!

services/web-app/utils/mdx-image-optimizer-plugin.mjs
jdharvey-ibm: image

Some food for thought for the URL stuff. The good part about the URL class is that you can modify the pieces individually and then output the result via `toString()`. So you can look at the hostname in isolation, make comparions/changes/etc. and do the same thing for the pathname separately without risking collision between the two.

alisonjoseph (dismissed):

alisonjoseph (approved):
👍👍

andreancardona (approved):
Looks good!

services/web-app/utils/mdx-image-optimizer-plugin.mjs
```diff
@@ -0,0 +1,104 @@
...
+        { type: 'mdxJsxAttribute', name: 'blurDataURL', value: base64 },
+        { type: 'mdxJsxAttribute', name: 'height', value: img?.height },
+        { type: 'mdxJsxAttribute', name: 'width', value: img?.width },
+        { type: 'mdxJsxAttribute', name: 'src', value: fullPath },
+        { type: 'mdxJsxAttribute', name: 'alt', value: node.alt }
+      ]
+      if (node.attributes && node.attributes.length) {
+        node.attributes.push(...nextImageAttributes)
+      } else {
+        node.attributes = nextImageAttributes
```
jdharvey-ibm, line 88: side note: this is an example of adding content into the AST without adjusting the offsets associated with the source material. The net result will be a different element at a new offset in the output, but the element is still able to render correctly. This type of approach should also work when wanting to push new child nodes into an AST.

services/web-app/utils/mdx-image-optimizer-plugin.mjs
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * @param {string} dirPath path where mdx source comes from (expected to be github repo url)
+ * @param {object} tree AST: mdx source
+ * @returns undefined
+ */
+const optimizeTreeImages = async (dirPath, tree) => {
+  const matches = []
+  visit(
+    tree,
+    (node) => node.type === 'image' || (node.type === 'mdxJsxFlowElement' && node.name === 'Image'),
```
jdharvey-ibm, line 53: It might be clearer to use a switch (or something similar) here to separate the two paths that are currently intertwined: one for if a markdown image is encountered and another if an Image mdxJsxFlowElement is encountered.

This would bring the logic more in line with typical XML parsers, whereby each unique type of element that is encountered has its own handler during the parse/visit.

services/web-app/utils/mdx-image-optimizer-plugin.mjs
jdharvey-ibm, line 88: side note: this is an example of adding content into the AST without adjusting the offsets associated with the source material. The net result will be a different element at a new offset in the output, but the element is still able to render correctly. This type of approach should also work when wanting to push new child nodes into an AST.

jharvey10 (changes requested):

francinelucca:
closing: we're starting over

PR #789, opened 2022-05-31
fix(web-app): shorten tag display names

### LINKED ISSUE #465: Shorten tag display names
Some of our component tags are long.

image

Same with our data viz collection tags.

image

In order to shorten these as tags in the filter popover, as well in the asset table per library:

image

Don't update the keys used in the schema, but update the display names to:

- [x] Contextual nav
- [x] Data viz
- [x] Structural nav
- [x] Geo overlay
- [x] Geo distortion

alisonjoseph:
Closes #465

#### Changelog

- Update tag display names

#### Testing / reviewing
Check the following tag display names are updated correctly.
Screen Shot 2022-05-31 at 8 59 36 AM


andreancardona (dismissed):
🌮  looks good to me!

francinelucca (commented):
library asset tags are not being shortened because they're displaying the keys instead of the names: 
image

Do we want to fix this here or open a ticket?

alisonjoseph:
Good catch, didn't even think to check that page. I can fix here.

francinelucca (approved):
✨

PR #796, opened 2022-05-31
fix(web-app): allow deep nesting in tree nav (styling)

### LINKED ISSUE #759: Sidenav active / open state needs to work with additional nesting
Need to support one more level deep of nesting in the left nav. It currently "works" but doesn't set an active class or keep the nav open. This will be for the Case studies section

francinelucca:
Closes #759 

*** Will remove test files once reviewed ***

#### Testing / reviewing

http://localhost:3000/test/test1
http://localhost:3000/test/test2
should look like this: 
image


alisonjoseph (dismissed):
LGTM!

alisonjoseph (approved):
🚀

andreancardona (approved):
💯

Issue #794, opened 2022-05-31
Create and publish markdown templates for all asset documentation pages


mattrosno:
Once asset documentation has been added to the schema (#792) and once that index content is wired up to conditionally render asset page tabs for usage, style, code, and accessibility (#793), we'll want markdown templates for each of those sub-pages so contributors can copy/paste the template so they have a skeleton of recommended headings and sections to fill out.

We have two component usage docs templates:

- https://carbondesignsystem.com/contributing/component/#template-for-components-with-one-variant
- https://carbondesignsystem.com/contributing/component/#template-for-components-with-multiple-variants

No style templates:

- https://carbondesignsystem.com/contributing/component/#style-guidance

One code template:

- https://carbondesignsystem.com/contributing/component/#code-guidance

No accessibility template.

And two pattern templates:

- https://carbondesignsystem.com/contributing/pattern/#how-to-write-a-pattern

**Acceptance criteria**

- [ ] The platform web app has contribution guidance where you can understand the indent of each asset documentation page, as well as templates to copy so we can guide users to have consistent asset documentation

mattrosno:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-Product-Management-624ca9397d28730018df40c2/issues/carbon-design-system/carbon-platform/794?planning-poker) with ZenHub @aubrey-oneal @kingtraceyj

Issue #485, opened 2022-04-04
Determine page tabs per asset


mattrosno:
Most all Carbon components are documented using page tabs for Usage, Style, Code, and Accessibility content. For example, the [Accordion](https://www.carbondesignsystem.com/components/accordion/usage). We are continuing to use sub pages for asset documentation given the length and maturity of existing content. This issue is to determine the naming of the page tabs going forward.

Another thing to keep in mind is our markdown templates for [components with one or many variants](https://www.carbondesignsystem.com/contributing/component/#template-for-components-with-one-variant) and [patterns with one or many variants](https://www.carbondesignsystem.com/contributing/pattern/#how-to-write-a-pattern).

Proposals in comments below.

**To do**

- [x] Mock up proposal 2 and 3 and make decision this week so we can make schema changes next sprint @kingtraceyj

mattrosno:
# Proposal 1

We simplify the page tabs to Overview, Design, Code and Accessibility.

### Overview

- What is it?
- When to use, when not to use it?
- What are related assets?
- Who maintains/contributes to it?

### Design

- How should the asset behave?
- What is the design spec?

### Code

- How to use the coded implementation?
- Where can I find additional API documentation?

### Accessibility

- What does provide?
- Design recommendations?
- Development considerations?

mattrosno:
Answering your questions @ljcarot:

>Is there a reason you are changing the tab labels?

Yes. Each page tab will need an entry in the schema. The schema is our opportunity to define what information lives where.

```yml
assets:
  accordion:
    name: Accordion
    status: stable
    thumbnailPath: /docs/accordion.svg
    overviewPath: /docs/overview.mdx
    designPath: /docs/design.mdx
    codePath: /docs/code.mdx
    accessibilityPath: /docs/accessibility.mdx
```

>Is a change needed before release 1.0?

Yes. Or else we won't know where to get that content per asset.

>Will changing the labels slow down the migration of content from CDS to Platform?

No. It's not the end of the world if some content is in the wrong tab during initial migration. That can get cleaned up over time.

kingtraceyj:
This is some initial thoughts— sorry a bit disorganized for your very organized issue. Again, we can chat in our design meeting but this is my initial reaction. 

### Tabs:
[...]

PR #805, opened 2022-06-01
feat(schemas): add assets docs

### LINKED ISSUE #792: Add asset page tabs to schema
Similar to https://github.com/carbon-design-system/carbon-platform/issues/555 for libraries, each asset in the schema needs keys that determine which content is available to fetch and render as page tabs. It could possibly be a `docs` or `documentation` object that belongs to each `asset`. For example,

```yml
assets:
  accordion:
    name: Accordion
    documentation:
      usagePath: https://github.com/carbon-design-system/carbon-website/blob/main/src/pages/components/accordion/usage.mdx
      stylePath: /docs/style.mdx
      codePath: /docs/code.mdx
      accessibilityPath: /docs/accessibility.mdx
```

**Acceptance criteria**

- [x] Asset schema is updated to support documentation
- [x] The intent of each asset documentation page is specified in the schema

**Dependents**

There will be follow-on issues to wire up indexed documentation paths to page tabs, and to provide markdown templates for all four documentation pages.

francinelucca:
Closes #792


docs/resource-schemas.md
```diff
@@ -45,6 +45,11 @@ assets:
       - data-display
     framework: react
     platform: web
+    docs:
+      usagePath: 'https://github.com/carbon-design-system/carbon-website/blob/main/src/pages/components/accordion/usage.mdx'
```
mattrosno, line 49: `docs` makes sense as the key. Other options: `documentation`, `pages`... since we can always add `navData` to assets later on if we need to add freeform page support to assets, I think this still makes sense to call these docs. It's asset documentation.

Also, I like how the docs keys end in `Path`, as that's consistent with the asset `thumbnailPath` and the library `packageJsonPath`.

docs/resource-schemas.md
```diff
@@ -205,6 +215,7 @@ assets:
...
 | `noIndex`         | If set to `true`, the global catalogs will exclude the asset.                                                                                            | Optional | Boolean          | No          | `false`       | –                                                                                                                                                                                                                                                                                                                                          |
+| `docs`            | Contains information on where to find supporting documentation for the asset. See [asset docs](#asset-docs).                                             | Optional | Object           | No          | -             | –                                                                                                                                                                                                                                                                                                                                          |
```
mattrosno, line 218:
```suggestion
| `docs`            | Contains information on where to find supporting documentation for the asset. See [asset docs](#asset-docs).                                             | Optional | Object           | Yes          | -             | –                                                                                                                                                                                                                                                                                                                                          |
```

This should be inheritable. For example, the component docs for all of our common components will live in the Carbon Styles library. That's how the ~5 libraries like Carbon React, Carbon Angular, etc. will get their asset documentation.

That sets us up to better keep code and docs in lockstep going forward. For example, let's say a component gets a new feature in Carbon Styles (meaning there are new styles and updated documentation.) Carbon React picks up that new Carbon Styles version and implements the new feature so the Carbon React inherits that updated documentation that describes the new feature. If Carbon Angular hasn't upgraded Carbon Styles yet and don't implement that new feature yet, the Carbon Angular docs will properly not have that new feature documented.

docs/resource-schemas.md
```diff
@@ -321,6 +332,37 @@ have the following values:
 | `cross-platform` | Runs natively on iOS, Android, and/or desktop. |
 | `web`            | Runs on the web.                               |
 
+#### Asset docs
+
+Asset docs is used to determine which supporting documentation is available to fetch and render as
+page tabs in the asset details for the asset as well as it's location. For each optional document
```
mattrosno, line 338:
```suggestion
page tabs in the asset details for the asset as well as its location. For each optional document
```

(Remove punctuation.)

docs/resource-schemas.md
```diff
@@ -321,6 +332,37 @@ have the following values:
...
+key, a path should be included; The path **must** be either a Github url according to the
+`https://{host}/{org}/{repo}/blob/{branch}/{path-to-file}` format, or a path relative to the
+carbon.yml file location.
+
+Currently supported document keys are:
+
+| docKey              | Description                                                                                                                                                                                       |
```
mattrosno, line 345: I'd replace `docKey` with `Document` since `docKey` looks like it'd be used in the YML.

docs/resource-schemas.md
```diff
@@ -321,6 +332,37 @@ have the following values:
 | `cross-platform` | Runs natively on iOS, Android, and/or desktop. |
 | `web`            | Runs on the web.                               |
 
+#### Asset docs
+
+Asset docs is used to determine which supporting documentation is available to fetch and render as
+page tabs in the asset details for the asset as well as it's location. For each optional document
+key, a path should be included; The path **must** be either a Github url according to the
```
mattrosno, line 339: Can you find-all replace `Github url` with `GitHub URL` from proper capitalization?

packages/schemas/carbon-resources.schema.json
```diff
@@ -147,6 +147,29 @@
                 "additionalProperties": false,
                 "required": ["type", "name", "url"]
               }
+            },
+            "docs": {
+              "description": "determines which supporting documentation is available to fetch and render as page tabs in asset details",
```
mattrosno, line 152:
```suggestion
              "description": "Determines which supporting documentation is available to fetch and render as page tabs in asset details.",
```

packages/schemas/internal/validation/asset-validation.schema.json
```diff
@@ -139,6 +139,29 @@
         "required": ["type", "name", "url"]
       }
     },
+    "docs": {
+      "description": "determines which supporting documentation is available to fetch and render as page tabs in asset details",
```
mattrosno, line 143:
```suggestion
      "description": "Determines which supporting documentation is available to fetch and render as page tabs in asset details.",
```

docs/resource-schemas.md
mattrosno: The schema.json files too.

mattrosno (commented):
@aubrey-oneal @kingtraceyj can you review these descriptions?

image

This came from our proposal 3 here: https://github.com/carbon-design-system/carbon-platform/issues/485#issuecomment-1089210709

If these initially look good, when we do https://github.com/carbon-design-system/carbon-platform/issues/794, we can revisit these definitions so we have a one-two sentence that describe the intent of each of the four page tabs (and their templates).

docs/resource-schemas.md
francinelucca: done!

docs/resource-schemas.md
francinelucca: done!

mattrosno:
Trying to simplify down to one-liners:

Usage - When to use the asset and how it behaves.
Style - Asset anatomy and design specifications.
Code - Asset API documentation (or where to find it.)
Accessibility - Asset accessibility considerations when designing and developing.

Thoughts @francinelucca @kingtraceyj @aubrey-oneal?

alisonjoseph:
@mattrosno am I mis-remembering that the first tab was going to be Overview not Usage? OR is the overview tab in addition to, not replacing Usage and not coming from a remote mdx file?

mattrosno:
> @mattrosno am I mis-remembering that the first tab was going to be Overview not Usage? OR is the overview tab in addition to, not replacing Usage and not coming from a remote mdx file?

@alisonjoseph correct first tab will always be Overview. That will only be structured data. The Overview will have dashboard, live demo, related assets, etc. That means we'll eventually have to trim some of that out of the existing Usage MDX files.

alisonjoseph:
@mattrosno gotcha, was wondering because of Linda's comment that the dot com components only have one page currently. https://www.ibm.com/standards/carbon/components/cards/ Ideally they would split that up and follow our templates, but in the interim will it be weird to have that all under Usage?

mattrosno:
@alisonjoseph in that scenario, we have options.

1. Don't add `overviewPath`. Index all of that content as `usagePath`. That means that Card component will have Overview and Usage tabs. Once we release platform v2 and we migrate all of Carbon for IBM.com content, we split that Card's usage content into other page tabs if need be.
2. Add `overviewPath` to schema and force all of that content on one overview page tab below the dashboard until content migration.

Let's confirm with @kingtraceyj. I'm guessing she'll want option (1) because two page tabs are better than one, and because the majority of that Card's content is usage guidance so it makes sense.

kingtraceyj:
@mattrosno As for the descriptions, they're looking good. I think the only one that could use a little more explanation is the style tab in the long version. We could add some clarity around what a design spec is. So maybe something more like this: Information on an assets design specs such as color, typography, and spacing.

And in the short description we could say Asset structure and design specifications. I’d take out the word anatomy since we have a section in usage docs that call out anatomy specifically.

alisonjoseph (approved):
🚀

docs/resource-schemas.md
```diff
@@ -321,6 +332,37 @@ have the following values:
...
+carbon.yml file location.
+
+Currently supported document keys are:
+
+| Document            | Description                                                       |
+| ------------------- | ----------------------------------------------------------------- |
+| `usagePath`         | When to use the asset and how it behaves.                         |
+| `stylePath`         | Asset structure and design specifications.                        |
+| `codePath`          | Asset API documentation (or where to find it).                    |
+| `accessibilityPath` | Asset accessibility considerations when designing and developing. |
```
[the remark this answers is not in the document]
francinelucca: Done!

mattrosno (approved):
The more detailed asset docs table looks great.

docs/resource-schemas.md
```diff
@@ -336,18 +336,16 @@ have the following values:
...
+| `stylePath`         | Asset structure and design specifications.                        | Optional | String | Yes         | -       | <ul><li>Github URL according to the `https://{host}/{org}/{repo}/blob/{branch}/{path-to-file}` format </li> <li>Path relative to the carbon.yml file location</li></ul> |
+| `codePath`          | Asset API documentation (or where to find it).                    | Optional | String | Yes         | -       | <ul><li>Github URL according to the `https://{host}/{org}/{repo}/blob/{branch}/{path-to-file}` format </li> <li>Path relative to the carbon.yml file location</li></ul> |
+| `accessibilityPath` | Asset accessibility considerations when designing and developing. | Optional | String | Yes         | -       | <ul><li>Github URL according to the `https://{host}/{org}/{repo}/blob/{branch}/{path-to-file}` format </li> <li>Path relative to the carbon.yml file location</li></ul> |
```
jdharvey-ibm, line 348: In order for these to be inheritable, they should have a default value. Otherwise, for the root-most asset in the inheritance chain, the value could be undefined. (since these are also marked as optional). Would it make sense for the default to be something like `./usage.mdx` (etc... for the others)?

This won't handle the case of a top-level carbon.yml defining all of the assets. @mattrosno Maybe that should push us in the direction of needing one carbon.yml per asset. Otherwise, we'll lose the association between an asset's source, its doc files, and its schema definition, which kind of goes against our co-location of assets and docs mentality.

mattrosno: @jdharvey-ibm here's an example written out that explains what we're after.

Carbon Styles library has Accordion, which has usage, style, code and accessibility documentation.

The Carbon React library inherits Carbon Styles. Carbon React also has an Accordion (that matches the Carbon Styles Accordion id.) Because the documentation pages are inherited, the Carbon React Accordion doesn’t need to specify anything as it will inherit the documentation pages from Carbon Styles.

That means the Carbon React Accordion doesn't need its own documentation. If inheriting the latest Carbon Styles version, the Carbon Styles Accordion usage page will be the same as the Carbon React Accordion usage page.

But, per our inheritance specification, it's possible to inherit a specific version to "pin" it. https://github.com/carbon-design-system/carbon-platform/blob/main/docs/resource-schemas.md#library-inheritance

So if Carbon Styles has a v1 and v2, and Carbon React is inheriting Carbon Styles v1, then when viewing the latest version of everything Carbon Styles Accordion will show v2 usage content and the Carbon React Accordion will show v1 usage content.

This becomes important when for example a feature was added (and documented) in Carbon Styles v2. If Carbon React hasn't implemented that new feature yet and is still inheriting v1, you don't want to see the documentation for that v2 feature in the Carbon React docs.

mattrosno: @jdharvey-ibm I typed out a lot there but glad that made sense.

Per https://github.com/carbon-design-system/carbon-platform/issues/485 (specifically proposal 3 that we landed on), if an asset doesn't specify any documentation pages, it only has the Overview tab. If it only has usage documentation for example, it only has Overview and Usage tabs.

If there's no code documentation, we won't want to be rendering a Code page tab with empty content.

Most of the PALs, and our universal patterns, only have usage documentation so a good number of our asset detail pages will have one or two page tabs for starters.

jdharvey-ibm: @mattrosno Ohhh okay got it. I was thinking that we'd be enforcing stuff like accessibility compliance by requiring those pages, but it sounds like that's not the case.

Given that, it seems acceptable to allow these to remain blank all the way up to the root-most components, but I'd still argue that we could throw some default values in there and try to fetch those during indexing. If they're there, awesome! We use them without any additional config from the user. If they don't, no problem! We just hide the tab.

mattrosno: I'm game for defaults and error handling that we'll need regardless.

Are you saying we support `usage.mdx`, `code.mdx`, `style.mdx`, `accessibility.mdx` where if they exist in the same directory, then we don't need to add `usagePath`, etc. to the indexing? If so, we can only support those defaults if we use `carbon.yml` for each asset?

That seems reasonable to me. Us using `carbon.yml` for an entire library seems like a short term thing. That file's going to get too big to manage. I think we'll eventually want a Carbon config file for the library and per asset in subdirectories.

francinelucca: I don't think defaults make sense if assets can be an array in our schema, are we making a decision to move back to having separate files/schemas for assets and libraries? if so then we can go with the defaults from my perspective

mattrosno: It'd be nice for us to keep support so a library can have one or many config files. I think it's reasonable to only support asset docs defaults if the Carbon config files are used on a per-asset basis?

jdharvey-ibm: We can definitely do both in the json schema.

https://json-schema.org/understanding-json-schema/structuring.html#ref

If the definition of an asset in the json schema is put into a "$def", it can be used in multiple places and extended. This is probably what we want to do with an asset if there's two different use cases for it in the schema. For example:

```json
"properties": {
  "myThing": {
    "$ref": "#/$defs/asset"
  },
  "myOverriddenThing": {
    "$ref": "#/$defs/asset",
    "properties": {
      "assetUsage": { "default": "wow a new default" }
    },
    "required": ["assetUsage"]
  }
}

"$defs": {
  "asset": {
    "type": "object",
    "properties": {
      "assetUsage": { "type": "string", "default": "usage.mdx" },
      "otherThing": { "type": "number", "default": 123 }
    },
    "required": ["assetUsage", "otherThing"]
  }
}
```

https://user-images.githubusercontent.com/51208233/172901257-6a9f724c-6294-4b45-895d-9fd3dc5bb03b.mov


So that could allow defaults to be used in one place and not another. We'd need a way to distinguish between the two though. For now, I think the defaults could exist in a multi-asset carbon.yml file and they won't hurt anything; they'll just point to files that don't exist, so we'd want to fall back to the inherited value or if none is found, don't show the tab.

Something to consider (outside the scope of this issue): are we moving away from an assets L0 (at least back-end-wise)? If so, it probably doesn't make sense for the schema definition to have an assets L0; it'd be better suited inside of the library object itself. That would enable us to do a top-level `asset` key for the one-carbon.yml-per-asset model, and `libraries.assets` as the multiple-assets-per-carbon.yml model, and then enforce that you only specify one or the other of those two.

francinelucca: ^^ sound good to me, added defaults! I think if we-re going the one-asset-per-carbon.yml way it'd probably make morse sense to have a separate carbon-asset.yml file like we used to with a separate asset schema instead of having a top-level `asset` ket on top of `library.assets` within the same file schema, do we need to open a separate ticket to discuss this further?

jharvey10 (approved):

docs/resource-schemas.md
jdharvey-ibm: Yeah, let's definitely open a discussion/investigation ticket for that to get it on the books.

mattrosno: @jdharvey-ibm @francinelucca a constraint to keep in mind (just a reminder) - it's common for some libraries to have a components folder where each component has its own JS file like `a-component.js` and `another-component.js` to both live in the `/components` directory.

We don't want to require libraries to move files around so it's "one asset per directory". That's why our current `carbon.yml` structure works well to handle this scenario of multiple assets in same directory.

francinelucca: Created #852 and copied some of these comments over 👯 . @mattrosno does this one look good to go?

mattrosno (approved):
@francinelucca yeah - looks good to me. The fun part will be the next issue to fetch those files, include page tabs, and page tab content. 🚀

docs/resource-schemas.md
jdharvey-ibm: @mattrosno In that situation you outlined, what's the typical strategy employed to distinguish the usage.mdx page for one component from another? Are they naming their mdx files in a way that somehow matches their component name?

mattrosno: I'm stating for a simple component that doesn't have a `my-component.module.scss` or a `my-component.stories.js` or tests or anything... if a component only lives in a single JS file, it's common for all of those components to live together in one directory.

In that scenario, we'd need to be able to index all of those components with one `carbon.yml` file.

The moment a component has a styles file, or stories, or tests, or documentation (like `usage.mdx`), at that point the component often lives in its own directory so it'd be natural to have the `carbon.yml` in that directory that only indexes that component.

PR #843, opened 2022-06-08
feat(web-app): add back to top button

### LINKED ISSUE #769: Add back to top button
Add back to top button, should function the same as the current Carbon website. Should be hidden on page load and display when you start scrolling down the page. 

Screen Shot 2022-05-23 at 1 12 33 PM

BackToTopBtn component
https://github.com/carbon-design-system/gatsby-theme-carbon/tree/main/packages/gatsby-theme-carbon/src/components/BackToTopBtn

Utils component (for scroll behavior example)
https://github.com/carbon-design-system/gatsby-theme-carbon/tree/main/packages/gatsby-theme-carbon/src/components/Utils

alisonjoseph:
Closes #769 

#### Changelog

**New**

- Back to top button

#### Testing / reviewing

Make sure back to top matches style and functionality on [Gatsby theme ](https://gatsby-theme-carbon.vercel.app/demo) (interior pages, Gatsby theme homepage is dark so the button tokens are different)

### DIFF
--- a/services/web-app/components/back-to-top/back-to-top.js
+++ b/services/web-app/components/back-to-top/back-to-top.js
@@ -21,12 +21,7 @@ const BackToTop = () => {
   const [hidden, setHidden] = useState(true)
 
   const scrollHandler = useCallback(() => {
-    if (window.scrollY >= 300) {
-      setHidden(false)
-    }
-    if (window.scrollY < 300) {
-      setHidden(true)
-    }
+    setHidden(window.scrollY < 300)
   }, [])
 
   useEventListener('scroll', scrollHandler)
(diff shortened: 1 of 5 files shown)


services/web-app/components/back-to-top/back-to-top.js
```diff
@@ -0,0 +1,49 @@
...
+const BackToTop = () => {
+  const [hidden, setHidden] = useState(true)
+
+  const scrollHandler = useCallback(() => {
+    if (window.scrollY >= 300) {
+      setHidden(false)
+    }
+    if (window.scrollY < 300) {
+      setHidden(true)
+    }
```
francinelucca, line 29:
```suggestion
    setHidden(window.scrollY < 300)
```
this will just set it to false if it doesn't meet the condition

francinelucca (commented):
small suggestion, I think wrapping the button in a scroller component or something is a good idea for the storybook solve 👍 . Another idea might be an "alwaysShow" prop or something that we pass on form storybook or a className 🤔

andreancardona (approved):
🎉 tested at different breakpoints looks good!

PR #854, opened 2022-06-09
feat: disallow indexing

### LINKED ISSUE #834: Disallow all using robots.txt until on TDL
As long as we're on a subdomain, let's add a robots.txt file to prevent any SEO collisions with carbondesignsystem.com. To do so, in the web-app public folder, a robots.txt with:

```
User-agent: *
Disallow: /
```

We'll have another issue for updating this when on the TLD.

alisonjoseph:
Closes #834

Add robots.txt file to public directory

### DIFF
--- a/services/web-app/public/robots.txt
+++ b/services/web-app/public/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /


services/web-app/public/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
+Disallow: /
```
francinelucca, line 2:
```suggestion
User-agent: *
Disallow: /

```
does this fix the no-newline thing?

francinelucca (approved):
💃

PR #867, opened 2022-06-10
refactor(web-app): programmatically add statusKey property to asset

### LINKED ISSUE #562: Add asset status utility function and clean up other places we're rendering asset status
When we pull library assets, we might want to do some transformation so that `asset.status` or some other property always has the asset status **key** (i.e., "stable"), even if status is just a value ("stable", "experimental"...), or an object ({key: "stable", note: "..."}). see note on: https://carbon-design-system.github.io/carbon-platform/resource-schemas.html#asset-status

Note: this includes fixing all places where we're doing something like
`const statusKey = asset?.content.status?.key ?? asset?.content.status ?? 'draft'` to just use the standardized property

francinelucca:
Closes #562

{{ Short description }}

#### Changelog

**New**

- `getAssetStatus` util
- updated places where asset status was being accessed directly to use statusKey instead

**Changed**

- Updated a bunch of cached files to have more up-to-date data which results in more variety of status 

#### Testing / reviewing

App should behave/look exactly the same, this is more of a code change.


services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js
```diff
@@ -201,9 +201,9 @@ const Asset = ({ libraryData }) => {
                       <dd className={dashboardStyles.meta}>
                         <StatusIcon
                           className={styles['status-icon']}
-                          status={assetData.content.status}
+                          status={assetData.statusKey}
                         />
-                        {get(status, `[${assetData.content.status}].name`, '–')}
+                        {get(status, `[${assetData.statusKey}].name`, '–')}
```
Comment (jdharvey-ibm) [inline services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:206]: I haven't seen this syntax before and had to look up what this does in the lodash docs. What's the advantage of doing this as opposed to `status[assetData.statusKey].name`?

services/web-app/utils/schema.js
```diff
@@ -20,11 +20,7 @@ import { type } from '@/data/type'
 export const statusSortComparator = (assetA, assetB) => {
   const statusKeys = Object.keys(status)
 
-  const statusA =
-    typeof assetA.content.status === 'object' ? assetA.content.status.key : assetA.content.status
-  const statusB =
-    typeof assetB.content.status === 'object' ? assetB.content.status.key : assetB.content.status
-  return statusKeys.indexOf(statusA) > statusKeys.indexOf(statusB) ? 1 : -1
+  return statusKeys.indexOf(assetA.statusKey) > statusKeys.indexOf(assetB.statusKey) ? 1 : -1
```
jdharvey-ibm, line 23: Unfortunately, this might not behave in the way you'd expect. The iteration order is always the same, but it's not necessarily the insertion order that this is relying on.

https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order

Also, if this gets transpiled to something ES2015 or older, there's no guarantee on the property order at all.

A better approach to achieve the intended behavior would be to store the status objects in an array where the `key` is one of the properties in each object, or in a `Map`.

You could also keep the object as-is, but have a second var in the `status.js` file which is an array of keys in the well-defined order.

services/web-app/utils/schema.js
```diff
@@ -76,6 +72,15 @@ export const getAssetType = (asset) => {
   return get(type, `${asset.content.type}`, type.component)
 }
 
+/**
+ * Gets the asset status string value. If not found, defaults to 'draft'.
+ * @param {import('../typedefs').Asset} asset
+ * @returns {string} Asset status value
+ */
+export const getAssetStatus = (asset) => {
+  return get(asset.content, ['status', 'key'], asset.content.status) ?? 'draft'
```
jdharvey-ibm, line 81: Same question here about the usage of lodash `get`. Is this the same is just doing this?:

```js
return asset.content.status.key || asset.content.status || 'draft'
```

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js
Comment (francinelucca) [reply services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:206]: in this particular case, that if the result of `status[assetData.statusKey].name` is undefined or null, it will default to '-' .

Also lodash won't error out if status[assetData.statuskey] is undefined and the other approach will

services/web-app/utils/schema.js
francinelucca: yeah same thing except in this case `...`.key could error out and lodash wont

services/web-app/utils/schema.js
francinelucca: added `assetStatusLifecycle` array

alisonjoseph (approved):

andreancardona (approved):
tested and works as expected!

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js
Comment (jdharvey-ibm) [reply services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/[asset].js:206]: Hmm. What about doing it with optional chaining to avoid the lodash overhead?

`status?.[assetData.statusKey]?.name || '-'`

services/web-app/utils/schema.js
francinelucca: done!

PR #876, opened 2022-06-13
refactor(web-app): separate useQueryState into two hooks

### LINKED ISSUE #425: web-app: useQueryState: separate responsibilities
See discussion on #313. 

**Summary**

- [x] Separate useQueryState into two, single-responsibility hooks:
  - useQueryState: only handles state through the query (non-persistent params, e.g.: page, pageCount, filter...)
  - useLocalState: only handles state through localStorage (persistent params, e.g.: view)

- [x] Revise usage of `useQueryState` across the app after refactor and move into `useLocalState` if necessary (probably only for `view` for now)

francinelucca:
Closes #425 

#### Changelog

**New**

- useLocalState hook

**Changed**

- useQueryState: only manages url query now, no local storage
- updated view query in catalog to use useLocalStorage instead

#### Testing / reviewing

App should behave more or less the same, except the term `view` shouldn't be present in the query anymore even if you change views (grid vs list)


services/web-app/utils/use-local-state.js
```diff
@@ -0,0 +1,86 @@
...
+        return
+      }
+
+      localStorage.setItem(`${router.pathname}:${key}`, JSON.stringify(newValue))
+      setValue(newValue)
+    },
+    [getValue, key, router.pathname, validateValue]
+  )
+
+  return [value ?? defaultValue ?? null, update]
```
jdharvey-ibm, line 83: The docs say this will return `''` if the defaultValue is not set. Should this match that? Or should the docs be updated to say that the default is null if you can't get the value and the defaultValue isn't set?

services/web-app/utils/use-local-state.js
```diff
@@ -0,0 +1,86 @@
...
+  key,
+  { defaultValue = '', parseNumbers = false, parseBooleans = false },
+  validateValue = () => true
+) => {
+  const router = useRouter()
+
+  const getValue = useCallback(() => {
+    if (typeof window === 'undefined') {
+      // localStorage is not available at this point
+      return null
```
jdharvey-ibm, line 43: See comment below about return type vs docs consistency

services/web-app/utils/use-query-state.js
```diff
@@ -25,75 +22,48 @@ const queryStringConfig = {
...
- * to the provided `defaultValue` every time a new page navigation to the path occurs;
- * supplying option `resetOnLoad` = `false` will cause the value to never be reset
- * 6- Supplying options `parseNumbers` or `parseBoolean` = `true` will cause the type of value
+ * the hook will return the defaultValue ('' if not set)
+ * 4- Supplying options `parseNumbers` or `parseBoolean` = `true` will cause the type of value
  * to be cast to desired type if possible, otherwise type will be string
  * @param {string} key Key to use in the query string
  * @param {{defaultValue: string, parseNumbers: boolean,
- * parseBooleans: boolean, resetOnLoad: boolean}} options Extra config options
+ * parseBooleans: boolean}} options Extra config options
```
jdharvey-ibm, line 32: Same comments on this function as above about return type consistency and param naming.

services/web-app/utils/use-local-state.js
francinelucca: maybe I just eliminate the ` ?? null` part? since defaultValue defaults to '' ? 
image

francinelucca: removed `defaultValue` default value (ha ha) and added extra clarification in docs to indicate undefined will be returned if defaultValue isn't provided (given that value is invalid)

services/web-app/utils/use-local-state.js
```diff
@@ -0,0 +1,86 @@
...
+ * 3 - The `validateValue` function receives the current query string value
+ * and should return true if the value is valid or false otherwise. If the value is invalid,
+ * the hook will return the defaultValue ('' if not set)
+ * 4 - Supplying options `parseNumbers` or `parseBoolean` = `true` will cause the type of value
+ * to be cast to desired type if possible, otherwise type will be string
+ * @param {string} key Key to use in the query string
+ * @param {{defaultValue: string, parseNumbers: boolean,
+ * parseBooleans: boolean}} options Extra config options
+ *
+ * @returns {{value, update}} Current value and update function
```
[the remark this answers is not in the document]
francinelucca: Done!

services/web-app/utils/use-query-state.js
francinelucca: Done!

services/web-app/utils/use-local-state.js
francinelucca: Done!

jdharvey-ibm (approved):
LGTM! 🚀

alisonjoseph (approved):

PR #878, opened 2022-06-13
refactor(web-app): starting to generalize assets catalog

### LINKED ISSUE #500: Refactor asset catalog components for search, filter, results, pagination to work in the libraries catalog
Visually, the libraries catalog and assets catalogs look the same for search, filter, results, empty states, and pagination.

The libraries page is currently just a list of links (basic styling). To make that more of a catalog when we have 40 or more libraries, we might need to refactor some of the asset components so they can be used here too.

francinelucca:
Closes #500 

{{ Short description }}

#### Changelog

**New**

- {{ New thing }}

**Changed**

- {{ Changed thing }}

**Removed**

- {{ Removed thing }}

#### Testing / reviewing

{{ Add descriptions, steps or a checklist for how reviewers can verify this PR works or not }}

### DIFF
--- a/services/web-app/components/assets-catalog/assets-catalog.js
+++ b/services/web-app/components/assets-catalog/assets-catalog.js
@@ -1,0 +1,59 @@
+/*
+ * Copyright IBM Corp. 2022, 2022
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import PropTypes from 'prop-types'
+import { libraryPropTypes } from 'types'
+
+import Catalog from '@/components/catalog'
+import CatalogItem from '@/components/catalog-item'
+import { getSlug } from '@/utils/slug'
+
+// import styles from './assets-catalog.module.scss'
+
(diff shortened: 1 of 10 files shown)


francinelucca:
> Screen Shot 2022-06-15 at 1 15 12 PM
> 
> Not sure whats going on here? 😅

hmm can you pull and try running it again? I had that issue in a previous rev but should be fixed 🤔

alisonjoseph (approved):
This is awesome! so excited for all the places we will be able to use this on the site 💯

francinelucca:
> On the live site currently Functions default to grid view, seems to all default to list view here, does that matter? Otherwise everything seems to be working perfectly. 👩‍🍳 💋

@alisonjoseph it's whatever you have saved in the localStorage, so if you're on Functions and you select grid and you reload, it should show as grid, same with list; but it's different localStorages for dev/test/prod so yeah it could be different depending on what you had previously saved.  (if nothing previously saved they all default to list, can test in incognito mode)

alisonjoseph:
@francinelucca ohh that makes much more sense. I was confused 🤣

andreancardona (approved):
Tested and looks good! 🔥

PR #899, opened 2022-06-17
fix(web-app): fix various library resources section styles

### LINKED ISSUE #578: Add resource links to library detail page
After talking with @aubrey-oneal - I confirmed that we don't need the body copy (D) in the design reference below for now.

**Inheritance**
- [x] Instead of showing "Depends on", let's keep it simple with "Inherits" if the library uses the `inherits` schema key
- [x] Only show the version if specified in the inherits key (see https://github.com/carbon-design-system/carbon-platform/blob/main/docs/resource-schemas.md#library-inheritance)
- [x] The library icon is available here and needs added it to the platform icons package https://ibm.ent.box.com/file/887827201496

**Source**
- [x] Instead of "Source", use "GitHub" or "IBM GitHub" to dual-purpose and the label and the host

**Package**
- [x] Only show if the package.json does not have `private: true` (see comment below)
- [x] The package icon is available here and needs added to the platform icons package https://ibm.ent.box.com/file/887827201496

**Design ref**
image

**Internal code packages**

We'll be indexing IBM Internal libraries that don't publish packages to npm. Instead, they often publish to artifactory or some other internal registry.

Looking at CloudPal for example, the package.json doesn't specify anywhere that it publishes to artifactory. If this package.json had `private: true` (as it should), then we know to not try to link out to npm. https://github.ibm.com/ibmcloud/pal/blob/master/patterns/package.json

francinelucca:
Closes #578 

#### Testing / reviewing

http://localhost:3000/assets/carbon-charts-react
Should match redline: https://ibm.box.com/s/s6btwm637udnek3pbjtfem412kgw83hp


services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/index.module.scss
```diff
@@ -35,3 +38,18 @@
...
+    color: theme.$text-secondary;
+  }
+
+  h4 {
+    @include type.type-style('body-compact-02');
+    @include mixins.line-clamp(2);
+
+    color: theme.$text-primary;
+  }
+}
```
Comment (alisonjoseph) [inline services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/index.module.scss:55]: This seems like it might be something we should add to the component. I'm going to check with design. https://ibm-studios.slack.com/archives/C022V4871K2/p1655499056464809?thread_ts=1655312718.200569&cid=C022V4871K2

services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/index.js
```diff
@@ -172,13 +177,14 @@ const Library = ({ libraryData, params, navData }) => {
...
         <Column sm={4} md={8} lg={8}>
-          <section>
+          <section className={styles['resources-section']}>
             <H2>Resources</H2>
 
             <CardGroup>
               {libraryData.content.inherits && libraryInheritanceCard()}
               <Column sm={4} md={4} lg={4}>
                 <ResourceCard
+                  aspectRatio={isLg && !isXlg ? '16:9' : '2:1'}
```
Comment (alisonjoseph) [inline services/web-app/pages/assets/[host]/[org]/[repo]/[library]/[ref]/index.js:187]: Also wondering if this should be part of the component.

jeanservaas:
As far as I can tell from the demo looking good and matches the spec.

kingtraceyj (approved):
from the screen recording, it's looking good!

services/web-app/components/resource-card/resource-card.js
```diff
@@ -41,8 +46,13 @@ const ResourceCard = (props) => {
     }
   )
 
+  // if aspectRatio is not specified and it's a card with title displaying at Lg breakpoint,
+  // default aspectRatio to 16:9, all other cases default to 2:1
```
alisonjoseph, line 50: Let's double check on this, I think we might want to default to 16:9 across the board at large, 2:1 at xlg.

alisonjoseph (approved):

PR #984, opened 2022-07-08
feat(web-app): rename sponsor to maintainer

### LINKED ISSUE #951: Replace Sponsor with Maintainer
Per @ljcarot's recommendation, update these Sponsor headings to Maintainer.

image

image

When community maintained, this can just say "Community".

image

Because `sponsor` isn't used anywhere in our schema, we can change our source code to use `maintainer` instead of `sponsor` too.

We'll need up update the filter too.

image

We'll still list individual contributors and maintainers on a per-library basis in a separate issue.

francinelucca:
Closes #951

#### Testing / reviewing

http://localhost:3000/libraries/carbon-components-svelte/latest
http://localhost:3000/libraries/carbon-components-svelte/latest/assets/button

### DIFF
--- a/services/web-app/components/asset-catalog-item/asset-catalog-item.js
+++ b/services/web-app/components/asset-catalog-item/asset-catalog-item.js
@@ -49,13 +49,13 @@ const AssetCatalogItemContent = ({ asset, isGrid = false, otherFrameworkCount = 
   const isLg = useMatchMedia(mediaQueries.lg)
 
   const { name, description } = asset.content
-  const { sponsor } = asset.params
+  const { maintainer } = asset.params
 
-  const sponsorTitle = teams[sponsor]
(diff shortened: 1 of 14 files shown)


alisonjoseph (changes requested):
I think we also might want to change sponsor->maintainer in the `libraries.mjs` file and in the filters?

We also use the word sponsor in a bunch of places in the app, like sponsorIcon etc. Do we want to do a find->replace for that too?

francinelucca:
> I think we also might want to change sponsor->maintainer in the `libraries.mjs` file and in the filters?
> 
> We also use the word sponsor in a bunch of places in the app, like sponsorIcon etc. Do we want to do a find->replace for that too?

you're right! changed all codebase from sponsor-> maintainer, there should be no references left to it now 💃

alisonjoseph (approved):
🚀

PR #1005, opened 2022-07-12
fix(web-app): allow  popover scrolling on mobile

### LINKED ISSUE #462: Popover needs set max height
### Browser

_No response_

### Description

Bug found by Matt - documenting here:
> @andreancardona @francinelucca ...we need to set a height on the popover content because when its parent is position sticky, and when the popover content is tall, you can't get to the bottom of the popover content.
> 
> 
> 

### Steps to reproduce

- [ ] change viewport size to mobile. In short view you cannot scroll to see the full content in the popover

https://user-images.githubusercontent.com/32720851/172722163-b0e486d0-3341-4def-9ff8-98d9d6224984.mov

alisonjoseph:
Closes #462

#### Changelog

**Changed**

- Add max height to popover and set overflow

#### Testing / reviewing

Ensure popover scrolling works on mobile, and other breakpoints are unchanged
http://localhost:3000/catalogs/components

### DIFF
--- a/services/web-app/components/catalog-multislect-filter/catalog-multiselect-filter.module.scss
+++ b/services/web-app/components/catalog-multislect-filter/catalog-multiselect-filter.module.scss
@@ -102,7 +102,11 @@
 .content {
   width: calc(100vw - var(--scrollbar-width));
   max-width: none;
+  max-height: 100vh;
+  overflow-y: auto;
+
   @include breakpoint.breakpoint(md) {
+    overflow: hidden;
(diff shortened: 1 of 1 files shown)


francinelucca (commented):
I don't seem to be able to scroll the full height 🤔

https://user-images.githubusercontent.com/40550942/178517300-38367bda-204f-4927-afb0-a9fc5731a313.mov

alisonjoseph:
@francinelucca I'll look at it again 😭

PR #1019, opened 2022-07-14
fix(web-app): // build warning

### LINKED ISSUE #871: Invalid href passed to next/router: //tab-one, repeated forward-slashes (//) or backslashes \ are not valid in the href
https://github.com/carbon-design-system/carbon-platform/runs/6841059709?check_suite_focus=true#step:6:270

I'm not sure which page is throwing this error/warning. I did a quick search in the repo for `//tab-one`, and 'tab-one', but didn't find anything.

The solution here is likely either to add a protocol to the front of the route (i.e. `https://example.com` instead of `//example.com`) or to check if `/` is being prepended to a path which already starts with `/`.

francinelucca:
Closes #871 

#### Changelog

**Changed**

- services/web-app/components/mdx-wrapper/mdx-wrapper.js : do not append baseSegment to tab route if empty

#### Testing / reviewing

- Nothing should've changed visually
- from `services/web-app` run `npm run build` and compare output against https://github.com/carbon-design-system/carbon-platform/runs/7246434865?check_suite_focus=true;  Verify lines 289-290,294-295 of step "Build, Push 🚀 "  are not present in local build
image

### DIFF
--- a/services/web-app/components/mdx-wrapper/mdx-wrapper.js
+++ b/services/web-app/components/mdx-wrapper/mdx-wrapper.js
@@ -47,9 +47,10 @@ const MdxWrapper = ({ children, ...props }) => {
         <PageTabs
           title={title}
           tabs={tabs.map((tab) => {
+            const tabSlug = slugify(tab, { strict: true, lower: true })
             return {
               name: tab,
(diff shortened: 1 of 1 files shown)


alisonjoseph (approved):
Confirmed the errors are gone when running npm run build locally

PR #1032, opened 2022-07-19
feat: error handling ui components

### LINKED ISSUE #995: MdxSanitizer - Create UI components
**Summary**

Designs: https://www.figma.com/file/WyfqQh9R1VXYZErfmsNDVH/MDX-errors?node-id=9%3A4035
In order to build the spec defined in #952 we'll need to introduce a couple new UI components:

- FullPageError
       - imports/exports
image

- InlineError (UnknownComponent)
       - UnknownComponent
       - parsing error
       - jsx error 
image
image

Can include code snippet:
image

These should be dynamic enough to handle different title/content as there can be multiple types of inline and fullPage errors.
Inlines should support link out as well for list of components or any other action

**Acceptance criteria**

What requirements must be met to complete this request?

1. New Components Are developed according to specs

andreancardona:
Closes: #995 

## TESTING: 
- test components here: http://localhost:3000/assets

TODO:

- [x] Fix CI checks linter issue
- [x] Remove 404.js before merging ⚠️


services/web-app/components/inline-notification/inline-notification.js
```diff
@@ -6,9 +6,16 @@
  */
 
 import { Column, Grid } from '@carbon/react'
-import { CheckmarkFilled, ErrorFilled, InformationFilled, WarningFilled } from '@carbon/react/icons'
+import {
```
francinelucca, line 9: we can probably revert all changes in this file, right?

services/web-app/components/inline-notification/inline-notification.module.scss
```diff
@@ -9,6 +9,10 @@
 @use '@carbon/react/scss/spacing' as spacing;
 @use '@carbon/react/scss/theme' as theme;
 
+:global(.cds--inline-notification__close-button) {
+  position: absolute;
+}
+
```
francinelucca, line 15: can remove this as well

services/web-app/components/page-not-found/page-not-found.module.scss
```diff
@@ -16,11 +16,3 @@
...
   margin: spacing.$spacing-10 0;
 }
-
-// override carbon max-width and padding-right
-.button {
-  width: 100%;
-  max-width: none;
-  padding-right: spacing.$spacing-05;
-  margin-top: spacing.$spacing-07;
-}
```
francinelucca, line 26: do we need to remove this?

services/web-app/components/full-page-error/full-page-error.js
```diff
@@ -0,0 +1,52 @@
...
+import styles from './full-page-error.module.scss'
+
+export const FullPageError = ({ title, subtitle, description, content }) => {
+  return (
+    <Grid>
+      <Column className={styles.column} sm={4} md={8} lg={{ start: 6, span: 6 }}>
+        <h1 className={styles.title}>{title}</h1>
+        <p className={styles.subtitle}>{subtitle}</p>
+        <p className={styles.description}>{description}</p>
+        <div className={styles.content}>{content}</div>
```
francinelucca, line 19: Do we need description and content here? I think we probably just need one of those and need to add an optional link at the bottom:
image
image

francinelucca (changes requested):
Looking good! some cleanup comms + suggestions

services/web-app/components/full-page-error/full-page-error.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+import { Column, Grid, Link } from '@carbon/react'
+import PropTypes from 'prop-types'
+
+import styles from './full-page-error.module.scss'
+
+export const FullPageError = ({ title, subtitle, content, link, href }) => {
+  return (
+    <Grid>
+      <Column className={styles.column} sm={4} md={8} lg={{ start: 6, span: 6 }}>
```
francinelucca, line 15: do we have an extra offset column here ? looks like figma is 4?
image

image

services/web-app/components/inline-error/inline-error.js
```diff
@@ -0,0 +1,99 @@
...
+    setShowNotification(!showNotification)
+  }
+
+  return (
+    showNotification && (
+      <Grid>
+        <Column sm={4} md={6} lg={12} className={clsx(styles.notification, className)}>
+          <div className={containerClassName}>
+            <Grid>
+              <Column sm={3} md={4} lg={9}>
```
francinelucca, line 36: let's confirm the inside grid specs here with design, I think 8/4 looks better? but the text spans to two lines . Figma also looks like it's 8/4 in the inner grid but it's really hard to read 🤔 
image
image

francinelucca (commented):
Some comms from grid division, otherwise LGTM! 💃 
should we get a design review on this? 👁️

services/web-app/components/full-page-error/full-page-error.js
andreancardona: > Some comms from grid division, otherwise LGTM! 💃 should we get a design review on this? 👁️

yes,  smart!

francinelucca (approved):
LGTM 👩‍🍳 💋

services/web-app/components/inline-error/inline-error.module.scss
```diff
@@ -0,0 +1,44 @@
...
+ * Copyright IBM Corp. 2022, 2022
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+@use '@carbon/react/scss/spacing' as spacing;
+@use '@carbon/react/scss/type' as type;
+
+:global(.cds--inline-notification__close-button) {
```
alisonjoseph, line 11: add override carbon comment

services/web-app/components/inline-error/inline-error.module.scss
```diff
@@ -0,0 +1,44 @@
...
+  position: absolute;
+}
+
+:global(.cds--inline-notification) {
```
alisonjoseph, line 15: add override carbon comment

alisonjoseph (approved):
LGTM other than adding a couple comments to the css

kingtraceyj:
I have a few questions. I see the grid conversation above, ideally the inline error will sit like this in the page and align with the content. is that what's happening currently?  
image

I'm also wondering if it would be possible to add in padding above the notification so that it doesn't sit up against the object above it. I would think 24 pixels would be a safe standard. 

Also, the inline notification doesn't need to be dismissible so we can hide the close button. I will update the redlines with that change.

andreancardona:
> I have a few questions. I see the grid conversation above, ideally the inline error will sit like this in the page and align with the content. is that what's happening currently? image
> 
> I'm also wondering if it would be possible to add in padding above the notification so that it doesn't sit up against the object above it. I would think 24 pixels would be a safe standard.
> 
> Also, the inline notification doesn't need to be dismissible so we can hide the close button. I will update the redlines with that change.
> 
> Would it also be possible to use Plex Mono for the inline code samples?

- [x] Yes -> Ideally the inline error will sit like this in the page and align with the content. is that what's happening currently?
- [x] I'm also wondering if it would be possible to add in padding above the notification so that it doesn't sit up against the object above it. I would think 24 pixels would be a safe standard.
- [x] Also, the inline notification doesn't need to be dismissible so we can hide the close button. I will update the redlines with that change.
- [x] Would it also be possible to use Plex Mono for the inline code samples?

alisonjoseph (commented):
Is there a reason we need to rebuild the inline error component and not just using/extending our mdx component that already exists? `<InlineNotification>` (with additional props)

or if not that, should we be using the Carbon component as a base? Or is there too much customization needed?

https://next.carbondesignsystem.com/mdx/inline-notification

andreancardona:
> hat, should we be using the

I was originally using the `InlineNotification` but started realizing it was more customization by adding the `x` close button, props and some styling - but now that the close button is removed. They are very similar. I could try just use the `InlineNotification`

kingtraceyj (changes requested):
looking at storybook, this looks like a carbon issue but is there a way we could do the action link 32px padding on the right? it looks like it's indented 64px right now and im guessing that's because of the close button.

Also, if there is code in the full page could we also use `code-02` like we do in the inline notification.
image

andreancardona:
> at storybook, this looks like a carbon issue but is there a way we could do the action link 32px padding on the right? it looks like it's indented 64px right now an

-[x] padding right 32px
-[x] updated font to `code-02`

kingtraceyj (approved):
looks good to go! 💥

francinelucca (changes requested):
nice refactor! ✨ ; should we make the inner components grid aware? looks like it's breaking on mobile 
image

alisonjoseph (commented):
LGTM! other than the small mobile changes we were chatting about on slack

francinelucca (approved):
❤️ it

PR #1033, opened 2022-07-21
fix(web-app): display design only tag in library assets

francinelucca:
Fixes logical error on library assets page that prevented Design only tag from showing up 

#### Changelog

**Changed**

- services/web-app/pages/libraries/[host]/[org]/[repo]/[library]/[ref]/assets/index.js: fix if statement condition

#### Testing / reviewing

http://localhost:3000/libraries/carbon-charts/latest/assets: validate design tags
image

### DIFF
--- a/services/web-app/data/type.js
+++ b/services/web-app/data/type.js
@@ -51,7 +51,7 @@ export const type = {
   },
   'design-only': {
     bgColor: gray[20],
-    name: 'Design Only',
+    name: 'Design only',
     path: '/',
     textColor: gray[70]
   },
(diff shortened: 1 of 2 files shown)


andreancardona (changes requested):
@francinelucca I think we might want to change the tag to `Design only` @aubrey-oneal can you confirm?

aubrey-oneal:
Unless it is a special vocab word, we'll always do sentence case.

So yes, it should be `Design only`

francinelucca:
fixed! thanks!! @aubrey-oneal @andreancardona ✨

andreancardona (approved):
😎

alisonjoseph (approved):

PR #1075, opened 2022-07-28
feat: storybook demo component

### LINKED ISSUE #1010: Create live Storybook demo MDX component
**Summary**

Design spec: https://www.figma.com/file/Cs94zXOUUEizXNdDz263OA/Component-demo?node-id=114%3A169975

**Functionality**

This can be achieved through a new MDX component with props to control embed width and the variants to include in the embed. The variants prop likely needs to be an array where each item has a display name and a URL.

The URL for each demo embed needs to be of this format so it's a "canvas-only" embed:

https://react.carbondesignsystem.com/iframe.html?id=components-button--default

Then, in the embed caption, the link to the full demo would take this format:

https://react.carbondesignsystem.com/?path=/story/components-button--default

(Replace `iframe.html?id=` with `?path=/story/`.)

The first demo in the array would get shown by default.

**Documentation**

We'll need to document how Storybook should be configured (e.g. use a theme switcher that puts `globals=theme:g10` as a query string parameter, component placement justification in a story, etc.)

Consider opening a separate issue to determine if we need to use an existing Carbon Storybook configuration package, or create a new one.

**Acceptance criteria**

- [ ] Render the caption that links out to the full demo
- [ ] Demo window sizes
  - [ ] 8 and 12 columns
  - [ ] Fixed height based on number of columns
- [ ] Theme switcher
  - [ ] Dropdown that updates `globals=theme:g10`
  - [ ] Background
- [ ] Variant switcher
  - [ ] Dropdown
  - [ ] Data bind
  - [ ] Edge cases for many variants

andreancardona:
Closes #1010 

This adds the `StorybookDemo` component

#### Testing / reviewing

- [x] http://localhost:3000/mdx/storybook-demo
- [x] Test the different props by changing the following: `size` `height` `themes` `variants`
- [x] Change different breakpoints
- [x] Checkout storybook

TODO:

- [x] border / psuedo element
- [x] fix default variant?

Thank you @francinelucca for help on some of the logic 🔥 🎉


services/web-app/components/storybook-demo/storybook-demo.js
```diff
@@ -0,0 +1,125 @@
...
+        )}
+        <Column sm={2} md={4}>
+          <Dropdown
+            titleText="Variant selector"
+            label="variant"
+            items={variants}
+            initialSelectedItem={initialSelectedVariant}
+            onChange={onVariantChange}
+          />
+        </Column>
```
francinelucca, line 84: I got the impression from the redlines that the variant selector was optional, I tried not sending the variants but the component errored out, are we covering for that scenario? 
image

services/web-app/components/storybook-demo/storybook-demo.js
```diff
@@ -0,0 +1,125 @@
...
+
+  const iframeUrl = url + '/iframe.html?id=' + variant + '&globals=theme:' + theme
+
+  const onVariantChange = (item) => {
+    setVariant(item.selectedItem.variant)
+  }
+
+  return (
+    <>
+      <Grid condensed>
```
francinelucca, line 64: md breakpoints looks weird to me 🤔 
image

francinelucca (dismissed):
This is looking GREAT ❤️ 
Question: did we decide against the component placement options? 
From Figma: 
image

services/web-app/components/storybook-demo/storybook-demo.js
alisonjoseph: +1 It should be 8 cols at medium

alisonjoseph:
@francinelucca I'm not sure we can control placement within the iframe, that will need to be controlled within the story itself I believe? We should double check. @andreancardona

services/web-app/components/storybook-demo/storybook-demo.js
alisonjoseph: Yea, was thinking about this one too. How would we want them to input that into the mdx if they only have one story. 

Would we want something like this
```
<StorybookDemo
  themeSelector={true}
  size="small"
  height="tall"
  url="https://react.carbondesignsystem.com/iframe.html?args=&id=components-button--default"
/>
```

or this, and if there is only one variant, then no switcher shows up?
```

<StorybookDemo
  themeSelector={true}
  size="small"
  height="tall"
  url="https://react.carbondesignsystem.com"
  variants={[
    {
      label: 'Button',
      variant: 'components-button--default'
    }
  ]}
/>
```

I'm leaning toward option 2.

francinelucca: we could support both? 👀

services/web-app/components/storybook-demo/storybook-demo.js
```diff
@@ -0,0 +1,125 @@
...
+            titleText="Variant selector"
+            label="variant"
+            items={variants}
+            initialSelectedItem={initialSelectedVariant}
+            onChange={onVariantChange}
+          />
+        </Column>
+      </Grid>
+      <Grid condensed>
+        <Column sm={4} md={columnSize} lg={columnSize} className={demoClassNames}>
```
alisonjoseph, line 87:
```suggestion
        <Column sm={4} md={8} lg={columnSize} className={demoClassNames}>
```

Fix for the medium breakpoint weirdness @francinelucca pointed out.

services/web-app/components/storybook-demo/storybook-demo.module.scss
```diff
@@ -0,0 +1,37 @@
...
+ * Copyright IBM Corp. 2022, 2022
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+@use '@carbon/react/scss/spacing' as spacing;
+@use '@carbon/react/scss/theme' as theme;
+
+:global(.cds--label) {
```
alisonjoseph, line 11:
```suggestion

// override carbon, label left padding
:global(.cds--label) {
```

services/web-app/components/storybook-demo/storybook-demo.module.scss
```diff
@@ -0,0 +1,37 @@
...
+  padding-left: spacing.$spacing-05;
+}
+
+:global(.cds--dropdown) {
```
alisonjoseph, line 15:
```suggestion

// override carbon, dropdown border
:global(.cds--dropdown) {
```

services/web-app/components/storybook-demo/storybook-demo.module.scss
```diff
@@ -0,0 +1,37 @@
...
+
+.storybook-demo {
+  height: 320px;
+  padding: spacing.$spacing-03;
+  background: theme.$layer-01;
+}
+
+.tall,
+.large {
+  height: 480px;
```
alisonjoseph, line 27:
```suggestion
  height: convert.rem(480px);
```

services/web-app/components/storybook-demo/storybook-demo.module.scss
```diff
@@ -0,0 +1,37 @@
...
+}
+
+.large.tall {
+  height: 640px;
```
alisonjoseph, line 31:
```suggestion
  height: convert.rem(640px);
```

services/web-app/components/storybook-demo/storybook-demo.module.scss
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright IBM Corp. 2022, 2022
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+@use '@carbon/react/scss/spacing' as spacing;
+@use '@carbon/react/scss/theme' as theme;
+
```
alisonjoseph, line 10:
```suggestion
@use '@carbon/react/scss/utilities/convert' as convert;

```

services/web-app/components/storybook-demo/storybook-demo.module.scss
```diff
@@ -0,0 +1,37 @@
...
+:global(.cds--label) {
+  padding-left: spacing.$spacing-05;
+}
+
+:global(.cds--dropdown) {
+  border-bottom: 1px solid theme.$layer-selected-01;
+}
+
+.storybook-demo {
+  height: 320px;
```
alisonjoseph, line 20:
```suggestion
  height: convert.rem(320px);
```

services/web-app/pages/mdx/storybook-demo.mdx
```diff
@@ -0,0 +1,42 @@
...
+    {
+      label: 'Set of buttons',
+      variant: 'components-button--set-of-buttons'
+    },
+    {
+      label: 'Skeleton',
+      variant: 'components-button--skeleton'
+    }
+  ]}
+/>
```
alisonjoseph, line 42:
```suggestion
## 8 Column / small

<StorybookDemo
  themeSelector={true}
  url="https://react.carbondesignsystem.com"
  variants={[
    {
      label: 'Button',
      variant: 'components-button--default'
    },
    {
      label: 'Secondary',
      variant: 'components-button--secondary'
    },
    {
      label: 'Tertiary',
      variant: 'components-button--tertiary'
    },
    {
      label: 'Ghost',
      variant: 'components-button--ghost'
    },
    {
      label: 'Danger',
      variant: 'components-button--danger'
    },
    {
      label: 'Icon button',
      variant: 'components-button--icon-button'
    },
    {
      label: 'Set of buttons',
      variant: 'components-button--set-of-buttons'
    },
    {
      label: 'Skeleton',
      variant: 'components-button--skeleton'
    }
  ]}
/>

## 8 Column / small -- tall

<StorybookDemo
  themeSelector={true}
  height="tall"
  url="https://react.carbondesignsystem.com"
  variants={[
    {
      label: 'Button',
      variant: 'components-button--default'
    },
    {
      label: 'Secondary',
      variant: 'components-button--secondary'
    }
  ]}
/>

## 12 Column / large

<StorybookDemo
  themeSelector={true}
  size="large"
  url="https://react.carbondesignsystem.com"
  variants={[
    {
      label: 'Button',
      variant: 'components-button--default'
    },
    {
      label: 'Secondary',
      variant: 'components-button--secondary'
    }
  ]}
/>

## 12 Column / large -- tall

<StorybookDemo
  themeSelector={true}
  size="large"
  height="tall"
  url="https://react.carbondesignsystem.com"
  variants={[
    {
      label: 'Button',
      variant: 'components-button--default'
    },
    {
      label: 'Secondary',
      variant: 'components-button--secondary'
    }
  ]}
/>

## No themeSelector

<StorybookDemo
  url="https://react.carbondesignsystem.com"
  variants={[
    {
      label: 'Button',
      variant: 'components-button--default'
    },
    {
      label: 'Secondary',
      variant: 'components-button--secondary'
    }
  ]}
/>

## No variants

<StorybookDemo
  url="https://react.carbondesignsystem.com"
  variants={[
    {
      label: 'Button',
      variant: 'components-button--default'
    }
  ]}
/>

```

Suggestion to show options on the page

alisonjoseph (changes requested):
So exciting seeing this working! 🥳 

Had a couple small comments, and added some mdx to show more examples (for demo / testing)
Screen Shot 2022-07-28 at 3 20 07 PM

services/web-app/components/storybook-demo/storybook-demo.js
```diff
@@ -0,0 +1,125 @@
...
+
+StorybookDemo.propTypes = {
+  /**
+   * Storybook demo height: default or tall
+   */
+  height: PropTypes.oneOf(['tall']),
+  /**
+   * Storybook demo size: small or large
+   */
+  size: PropTypes.oneOf(['small', 'large']),
```
alisonjoseph, line 110: Do we like this prop name? `size` 

It feels a little confusing to me, but not sure what would be better. `width`? or just add a `wide` prop instead?

What do you think @francinelucca @andreancardona

francinelucca: ohh I like `wide`, I also thought  `height: 'tall'` was weird if there's only one option, maybe we can do just `wide` and `tall`?

alisonjoseph: ooh yea, I like that 👍

kingtraceyj (changes requested):
it's looking good. only a small visual tweak from me. there should be line in between the two drop downs `$border-subtle`

image

i'm going to send some screen captures over to anna just to make sure i didn't miss anything.

services/web-app/components/storybook-demo/storybook-demo.js
mattrosno: Any convention from the system squad that we could use? Standardizing props was a big v11 focus.

alisonjoseph: Good point Matt, I believe the convention they landed on is
`size` `'xs''sm''md''lg''xl'` (some components don't have all 5 sizes)

However would size apply to the width or the height for us, that's where this component gets a bit confusing, and wide/tall might make more sense than size?

alisonjoseph: Checked w/ devs in slack, sounds like there aren't any components with width/height options https://ibm-studios.slack.com/archives/GCQ5R263T/p1659103062541399 so should be good to move forward with `wide` and `tall` props @andreancardona

andreancardona:
> 

fixed!

andreancardona:
> it's looking good. only a small visual tweak from me. there should be line in between the two drop downs `$border-subtle`
> 
> image
> 
> i'm going to send some screen captures over to anna just to make sure i didn't miss anything.

fixed!

kingtraceyj (changes requested):
it's looking like there is now a space and a line between the dropdown items. you can remove the space.

image

alisonjoseph (changes requested):
Looks like it's selecting the second variant not the first as the default now? Are you seeing this too?

services/web-app/components/storybook-demo/storybook-demo.module.scss
```diff
@@ -0,0 +1,45 @@
...
+  padding-left: spacing.$spacing-05;
+}
+
+:global(.cds--dropdown) {
+  border-bottom: 1px solid theme.$layer-selected-01;
+}
+
+.theme-selector {
+  :global(.cds--dropdown) {
+    border-right: 1px solid theme.$layer-selected;
```
alisonjoseph, line 23: Instead of a border I would use the :after pseudo element with a width of 1px to fill the space that is already there.

alisonjoseph:
Just realized we need the disclaimer at the bottom with the full storybook demo link 😄

andreancardona:
> Just realized we need the disclaimer at the bottom with the full storybook demo link 😄

added it!

services/web-app/components/storybook-demo/storybook-demo.js
```diff
@@ -0,0 +1,144 @@
...
+            allowtransparency="true"
+            allowFullScreen="true"
+          />
+        </Column>
+      </Grid>
+      <Grid>
+        <Column sm={4} md={7}>
+          <p className={styles.disclaimer}>
+            This live demo contains only a preview of functionality and styles available for this component.
+            View the <Link className={styles['disclaimer-link']} href="https://next.carbondesignsystem.com/">full demo</Link> on Storybook
```
alisonjoseph, line 112: This link should be similar to the embed url but Replace `iframe.html?id=` with `?path=/story/`

Which makes me think we don't support this, and require the variant to be added to the variants prop. 
```

<StorybookDemo
  themeSelector={true}
  size="small"
  height="tall"
  url="https://react.carbondesignsystem.com/iframe.html?args=&id=components-button--default"
/>

```

alisonjoseph: @francinelucca @alisonjoseph

services/web-app/components/storybook-demo/storybook-demo.js
```diff
@@ -0,0 +1,144 @@
...
+            src={iframeUrl}
+            frameBorder="no"
+            allowtransparency="true"
+            allowFullScreen="true"
+          />
+        </Column>
+      </Grid>
+      <Grid>
+        <Column sm={4} md={7}>
+          <p className={styles.disclaimer}>
```
alisonjoseph, line 110: I think we can just use the `<Caption> `component here vs custom styles

services/web-app/components/storybook-demo/storybook-demo.js
andreancardona: yep agreed!

andreancardona:
@alisonjoseph @kingtraceyj @francinelucca This is ready for review!  :)

services/web-app/components/storybook-demo/storybook-demo.js
```diff
@@ -0,0 +1,151 @@
...
+    [styles.tall]: tall,
+    [styles.wide]: wide
+  })
+
+  const [theme, setTheme] = useState(themeItems[0].src)
+  const onThemeChange = (item) => {
+    setTheme(item.selectedItem.src)
+  }
+
+  const multipleVariants = variants.length > 1
```
francinelucca, line 53:
```suggestion
  const multipleVariants = variants?.length > 1
```

services/web-app/components/storybook-demo/storybook-demo.js
```diff
@@ -0,0 +1,151 @@
...
+          </Column>
+        )}
+        {variantsDefined && multipleVariants && (
+          <Column sm={2} md={4}>
+            <Dropdown
+              id="variant-selector"
+              titleText="Variant selector"
+              label="variant"
+              items={variants}
+              initialSelectedItem={variants[0].label}
```
francinelucca, line 95:
```suggestion
              initialSelectedItem={variants[0]?.label}
```

services/web-app/components/storybook-demo/storybook-demo.js
```diff
@@ -0,0 +1,151 @@
...
+   */
+  themeSelector: PropTypes.bool,
+  /**
+   * Storybook demo url to change themes and variants
+   */
+  url: PropTypes.string,
+  /**
+   * Storybook demo variants for the specified component
+   */
+  variants: PropTypes.object,
```
francinelucca, line 144:
```suggestion
  variants: PropTypes.object.isRequired,
```

services/web-app/components/storybook-demo/storybook-demo.js
```diff
@@ -0,0 +1,151 @@
...
+   */
+  tall: PropTypes.bool,
+  /**
+   * Storybook demo display or hide theme selector
+   */
+  themeSelector: PropTypes.bool,
+  /**
+   * Storybook demo url to change themes and variants
+   */
+  url: PropTypes.string,
```
francinelucca, line 140:
```suggestion
  url: PropTypes.string.isRequired,
```

francinelucca (changes requested):
LGTM! minor suggestions

kingtraceyj (changes requested):
this is so small—there is one pixel missing below the line between the dropdowns.

image

andreancardona:
> this is so small—there is one pixel missing below the line between the dropdowns.
> 
> image

good catch! Fixed :)

francinelucca (approved):
💃

alisonjoseph (approved):
🔥 SO GOOD! 🔥 

I pulled in latest from main, and pushed a tiny fix for the storybook story. 😄

PR #1111, opened 2022-08-02
feat(web-app): side nav ia updates

### LINKED ISSUE #1037: Update side nav items
**Design**

- [x] Specify page header background color and pictogram for Libraries and Design kits catalogs, if those are to not be black with no pictogram

**Dev**

- [x] Move Accessibility and Content into a Guidelines item beneath Elements
image

- [x] Move Libraries into the Catalogs section and put Components, Functions, Patterns, Templates under Assets
image

- [x] Update the page header to "Data visualization" (no capital V)
image

- [x] Move FAQ and Content as the last two items in About Carbon section
image

- [x] Update "Asset catalog" to "Asset catalogs" (plural) on the home page
image

alisonjoseph:
Closes #1037

#### Testing / reviewing

Check that side nav matches #1037 and that all links work correctly

@aubrey-oneal review the updates urls to match the new side nav and make sure its ok. Dev team discussed in length, but would like your opinion / sign off.

http://localhost:3000/catalogs/libraries/carbon-react/latest/assets/accordion

http://localhost:3000/data-visualization

http://localhost:3000/assets/components
http://localhost:3000/assets/functions
http://localhost:3000/assets/patterns
http://localhost:3000/assets/templates

Screen Shot 2022-08-04 at 11 37 10 AM


francinelucca (approved):
💃 💃

mattrosno:
@alisonjoseph no opinion here if we update our asset catalog URLs to have `/assets`.

alisonjoseph:
@mattrosno that also means we will end up with assets twice in the url for the detail pages, so maybe we leave as is, without complicating the url further.

mattrosno:
@alisonjoseph detail pages are under `/libraries` route, not `/catalogs` route, so we should be good there. Since catalogs are shortcuts to assets that live in libraries.

francinelucca:
on that same note, should `/libraries` be `/catalogs/libraries` now? since it's insider the catalog? 👀 
I vote yes...

mattrosno:
@alisonjoseph I think we landed on:

https://next.carbondesignsystem.com/assets/components
https://next.carbondesignsystem.com/libraries
https://next.carbondesignsystem.com/data-visualization

Is that right?

alisonjoseph:
@mattrosno correct, removing "catalogs" and "collections" from urls

mattrosno:
Logic: remove `/catalogs` and `/collections` from URLs because those are just headings in the side nav (not clickable.) If we need to add in `/collections` in the future to protect against URL collisions when those are user defined, we can, and would only have to add redirects for the data viz collection.

mattrosno:
@alisonjoseph and adding `/assets` in the asset catalog routes still? https://next.carbondesignsystem.com/assets/components

francinelucca (approved):
💃

andreancardona (approved):
🔥

services/web-app/data/nav-data.js
```diff
@@ -231,43 +249,35 @@ export const assetsNavData = [
         title: 'Data vizualization',
```
aubrey-oneal, line 249: change to 'visualization'

aubrey-oneal (changes requested):
URLs look good! Just need to change 'vizualization' to 'visualization'

PR #1127, opened 2022-08-10
feat: render remote asset thumbnail SVGs

### LINKED ISSUE #297: Use remote thumbnails instead of local id matching
Right now, all of the thumbnails are committed to the platform web-app service and we match filename and asset ids to determine when to show images. That was enough to prototype and hit the private beta, but we need a more scalable approach that doesn't have id conflicts and allows thumbnail updating from the remote repositories.

## Option 1 

- Document via the schema that thumbnail images must be SVG
- Continue to use the `thumbnailPath` schema key
- As we fetch data for each remote GitHub repository per library, we read the content for the SVG files and run it through SVGO to optimize
- Either use SVGO to convert to React components, or leverage Next.js Image component, or inline the SVG

## Option 2

- For now require all thumbnail images to be in the platform repo
- Place images in a folder per library and still match by filename and asset id to prevent collisions

## Option 3

Or... don't do anything until we have our image service in place...

mattrosno:
Closes #297 

This PR implements option 1 detailed in the linked issue, detailed back in February and March.

This PR adds a new property to the asset schema `thumbnailSvg`. It's like a private property where we don't need to document it as we expect `thumbnailPath` to be set. If `thumbnailPath` is provided, this new logic:

1. Validates and resolves the path
2. Gets the content of that file
3. Optimizes the SVG
4. Saves the optimized SVG string to `thumbnailSvg` in the data model

This comes with the benefits:

- Replaces thumbnail matching by `id`s
  - Helps us find indexing bugs now (we weren't previously using `thumbnailPath`s anywhere)
  - Allows asset maintainers to see the thumbnails they index (without learning our shortcut of `id` matching from this repo, where there can be `id` conflicts, etc.)
  - Makes it faster for us to index new assets because we don't have to save the SVGs in multiple repos
- Optimizes SVGs using SVGO resulting in smaller SVG file sizes
- Inlines SVGs (see screenshot below) so the thumbnails are render on initial page load (no waiting on additional network requests for the SVGs)
- Caches the optimized SVGs instead of the originals (SVGO only runs each time there's a cache miss and the SVG files are fetched from GitHub; the SVGs aren't optimized every time they are needed by the web app)

#### Changelog

**New**

- `svgo` dependency for web app, MIT license
- `thumbnailSvg` property in the asset schema
- `file-cache.js` logging component
- `getSvgResponse` function in the file cache, which attempts to optimize the SVG then cache the optimized SVG instead of the original
- `getThumbnailPath` in `/lib/github.js` to validate the `thumbnailPath` before requesting its contents
- Logic to find all thumbnail paths when getting a library's assets, and then fetch and optimize all of those SVGs, and then merge the optimized SVG into the asset data model as `thumbnailSvg`

**Changed**

- Asset catalog item thumbnails now render as inline SVG instead of SVG files that previously required additional network requests, so there's no delay in page load and seeing the thumbnails
- I reset the dev libraries in `/data/libraries.mjs` to match production, and cleared the development data in `/public/data` to validate these changes

**Removed**

- All of the thumbnails in `/public/assets/thumbnails` except the fallback when there is no thumbnail

#### Testing / reviewing

Everything should look as it did before. See comment below where the logging identified 5 thumbnail paths that need fixed.


mattrosno:
4 thumbnails to fix:

```
8/10/2022, 7:22:12 PM service:local-dev info [file-cache.js] Unable to optimize the SVG from github.com/carbon-design-system/carbon-components-svelte/thumbnails/inline-loading.svg with the ref master
8/10/2022, 7:22:12 PM service:local-dev info [file-cache.js] Unable to optimize the SVG from github.com/IBM/carbon-components-angular/src/thumbnails.tooltip-icon.svg with the ref master
8/10/2022, 7:22:12 PM service:local-dev info [file-cache.js] Unable to optimize the SVG from github.com/carbon-design-system/carbon-charts/thumbnails/histogam.svg with the ref master
8/10/2022, 7:22:12 PM service:local-dev info [file-cache.js] Unable to optimize the SVG from github.com/carbon-design-system/ibm-cloud-cognitive/packages/security/thumbnails/filter-panel.svg with the ref main
8/10/2022, 7:22:12 PM service:local-dev info [file-cache.js] Unable to optimize the SVG from github.com/carbon-design-system/carbon-for-ibm-dotcom/thumbnails/leads-pace-search.svg with the ref main
```

mattrosno:
Before on left, after on right, shows how SVGs are now inlined meaning there's no additional network request per thumbnail, so there's no rendering delay per thumbnail.
image

services/web-app/lib/file-cache.js
```diff
@@ -96,6 +100,40 @@ export const getResponse = (host, route, options) => {
...
+      }
+    } catch (err) {
+      logging.info(
+        `Unable to optimize the SVG from ${host}/${options.owner}/${options.repo}/${options.path} with the ref ${options.ref}`
+      )
+
+      return data
+    }
+  })
+}
```
francinelucca, line 135: I think we need to await the `_getresponse` otherwise the data might not have been resolved by the time the buffer is created, tested deleting the cached files and it wouldn't work for me until I added the await 
```suggestion
export const getSvgResponse = async (host, route, options = {}) => {
  const responseKey = slugifyRequest(host, route, options)
  // console.log('CACHE HIT', responseKey)

  return diskCache.wrap(responseKey, async () => {
    const data = await _getResponse(host, route, options)

    try {
      const originalSvgString = Buffer.from(data.content, data.encoding).toString('utf8')
      const optimizedSvgResult = optimize(originalSvgString)
      const optimizedSvgString = optimizedSvgResult.data

      return {
        ...data,
        content: Buffer.from(optimizedSvgString, 'utf8').toString('base64')
      }
    } catch (err) {
      logging.info(
        `Unable to optimize the SVG from ${host}/${options.owner}/${options.repo}/${options.path} with the ref ${options.ref}`
      )

      return data
    }
  })
}
```

francinelucca (changes requested):
❤️  ittt , some CRs

services/web-app/lib/file-cache.js
mattrosno: I originally had an await there too, but Sonarlint threw an error, saying the await was unnecessary?

packages/schemas/carbon-resources.schema.json
```diff
@@ -108,6 +108,10 @@
               "type": "string",
               "description": "Relative location of the asset's thumbnail image."
             },
+            "thumbnailSvg": {
+              "type": "string",
+              "description": "Asset's thumbnail image as an SVG string."
+            },
```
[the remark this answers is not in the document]
mattrosno: Thanks for the review - will do!

services/web-app/lib/file-cache.js
francinelucca: probably something to do with the return type of `_getRepsonse` : 
```
/**
 * Internal function that proxies GitHub requests.
 * @param {string} host - GitHub API base URL
 * @param {string} route - GitHub API request route
 * @param {object} options - Options merged into the request route
 * @returns {object} GitHub API response data
 */
```
try making it a promise instead: 
```
/**
 * Internal function that proxies GitHub requests.
 * @param {string} host - GitHub API base URL
 * @param {string} route - GitHub API request route
 * @param {object} options - Options merged into the request route
 * @returns {Promise<object>} GitHub API response data
 */
```
I can help troubleshoot if that still doesn't fix it

mattrosno: Ah. I didn't even think of invalid JSDoc throwing off SonarLint but that makes sense. I'll make another pass over this to make sure.

mattrosno:
@francinelucca I added the async's and await back in, and updated that JSDoc comment, and SonarLint passes now. Also, in that last commit, I cleared all dev data and committed that fresh, so everything appears to still be working from my end.

@carbon-design-system/carbon-platform-devs this is ready for reviews. Thanks!

francinelucca (approved):
working for me now!

alisonjoseph (approved):
LGTM! 


Screen Shot 2022-08-17 at 2 48 13 PM 🔥

PR #1128, opened 2022-08-10
feat(web-app): add related libraries to dashboard

### LINKED ISSUE #1039: Add related libraries to library dashboard
Redline: https://www.figma.com/file/K6HobXD4KXUIdkLrVjUzwK/LibraryDetail_CarbonReact?node-id=0%3A1

- [x] If a library is part of a group (defined in `/web-app/data/libraries.mjs`, comma-separate the other libraries and show the display name of each
- [x] Don't include a library if it's `noIndex: true`, because that would be shown below in the Resources section if its inherited
- [x] If the library isn't part of a group, show `–` (just like how we handle assets dashboard if there are no "Tags")

This is so we can cross-link libraries that (should) have the same components, but build with different JavaScript libraries.

See red highlight:
image

alisonjoseph:
Closes #1039 

#### Changelog

**New**

- add related libraries to library dashboard
- add correct right padding to dashboard items

#### Testing / reviewing

Check that related library links are correct and working
http://localhost:3000/libraries/carbon-charts-angular

### DIFF
--- a/services/web-app/pages/libraries/[host]/[org]/[repo]/[library]/[ref]/index.js
+++ b/services/web-app/pages/libraries/[host]/[org]/[repo]/[library]/[ref]/index.js
@@ -263,22 +263,22 @@ export const getServerSideProps = async ({ params }) => {
   }
 
   const relatedLibs = []
-if(libraryData.params.group){
-  for (const [slug, libraryParams] of Object.entries(libraryAllowList)) {
-    if (libraryParams.group === libraryData.params.group) {
-      const relatedLibData = await getLibraryData({
-        library: slug,
-        ref: 'latest',
-        ...libraryParams
-      })
-      if (
-        relatedLibData?.content.id !== libraryData.content.id &&
-        !relatedLibData?.content?.noIndex
-      ) {
-        relatedLibs.push(relatedLibData)
+  if (libraryData.params.group) {
+    for (const [slug, libraryParams] of Object.entries(libraryAllowList)) {
+      if (libraryParams.group === libraryData.params.group) {
+        const relatedLibData = await getLibraryData({
+          library: slug,
+          ref: 'latest',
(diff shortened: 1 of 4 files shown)


services/web-app/pages/libraries/[host]/[org]/[repo]/[library]/[ref]/index.js
```diff
@@ -246,6 +262,25 @@ export const getServerSideProps = async ({ params }) => {
...
+        ...libraryParams
+      })
+      if (
+        relatedLibData?.content.id !== libraryData.content.id &&
+        !relatedLibData?.content?.noIndex
+      ) {
+        relatedLibs.push(relatedLibData)
+      }
+    }
+  }
```
Comment (francinelucca) [inline services/web-app/pages/libraries/[host]/[org]/[repo]/[library]/[ref]/index.js:280]: My bad, we need to assert that the current library has a group before finding the related libs, otherwise it's matching libraries that have `group: undefined`, example: http://localhost:3000/libraries/ibm-security
so just need to wrap the hole thing in a `if(libraryData.params.group)`; probably need to do the same thing in the otherAssetFrameworks logic too 👉 👈 
```suggestion
if(libraryData.params.group){
  for (const [slug, libraryParams] of Object.entries(libraryAllowList)) {
    if (libraryParams.group === libraryData.params.group) {
      const relatedLibData = await getLibraryData({
        library: slug,
        ref: 'latest',
        ...libraryParams
      })
      if (
        relatedLibData?.content.id !== libraryData.content.id &&
        !relatedLibData?.content?.noIndex
      ) {
        relatedLibs.push(relatedLibData)
      }
    }
  }
  }
```

francinelucca (changes requested):
LGTM! added CR for something we overlooked in the related lib logic

francinelucca (approved):

andreancardona (approved):
tested and looks good!

PR #1146, opened 2022-08-17
feat: add page header colors and pictograms

### LINKED ISSUE #1056: Library and design kit catalog colors
Now that libraries and design kits have moved to the catalog section, they will need unique colors to differentiate them from other assets.

mattrosno:
Closes #1056 

This PR updates all page header backgrounds and pictograms according to the updated design spec. The design kits catalog has its own page header. All pages in a library (except assets) have a library page header. All pages in a collection have a collection page header.

The PR also removes a couple pages not needed for v1, and adds placeholder design kit pages that are needed for v1.

I didn't mean to commit the `/public/data` JSON files. Is that okay?

#### Changelog

**New**

- `/data/page-headers.js` that extend `/data/asset-types.js`, and specify page header background colors and pictograms for collection, design kit, and library pages
- `pageHeaderType` prop in the `RemoteMdxLoader`, so you can specify if the page should have a collection, design kit, or library page header
- `pageHeaderType` frontmatter support in `MdxWrapper` so *.mdx files can have a collection, design kit, or library page header
- Design kits catalog placeholder page
- Library design kits placeholder page

**Changed**

- Renamed `/data/types.js` to `/data/asset-types.js` and cleaned it up to be only colors, icons, and display text for asset types
- This meant I had to remove the `design-only` object in there, since that's not an asset type, but a framework key
- After removing that, I refactored the `TypeTag` so you can pass in a name, and so the default colors are gray if no type is passed in
- All pages in the Data viz collection now have the collection page header
- Libraries catalog get the library page header
- All pages in a library (assets, design kits, tutorial pages) get the library page header, except assets that remain unchanged

**Removed**

- Commented out code for an elements catalog, since that's post-v1
- Elements catalog page because it's not getting used, and is post-v1
- Versions nav data item for each library, because that's post-v1

#### Testing / reviewing

1. Design only tags here look as they did before http://localhost:3000/libraries/carbon-charts/latest/assets


kingtraceyj (approved):
looks good to go! 💥

alisonjoseph (approved):
LGTM! Thanks Matt

PR #1149, opened 2022-08-17
fix(web-app): sass import order

alisonjoseph:
Fix import order for sass 

#### Testing / reviewing

Nothing should have changed


francinelucca (approved):
confirmed app still working as expected

andreancardona (approved):
everything still looks good 🔥

PR #1159, opened 2022-08-19
fix(web-app): hide view asset doc button if no external docs

### LINKED ISSUE #1141: "View asset docs" button is always showing
**Description**

It looks like the "View asset docs" button is showing, even if an asset doesn't have external docs.

**Steps to reproduce**

1. Go here: https://next.carbondesignsystem.com/libraries/carbon-react/latest/assets/use-theme
2. It shouldn't have that secondary button, because `use-theme` asset doesn't have `externalDocsUrl`: https://github.com/carbon-design-system/carbon/blob/main/packages/react/carbon.yml#L693

image

alisonjoseph:
Closes #1141

Only show view asset docs button if link exists

#### Testing / reviewing

Test asset with and without asset docs

http://localhost:3000/libraries/carbon-react/latest/assets/use-context-menu
http://localhost:3000/libraries/carbon-react/latest/assets/accordion


services/web-app/pages/libraries/[host]/[org]/[repo]/[library]/[ref]/assets/[asset]/index.js
```diff
@@ -278,14 +278,16 @@ const Asset = ({ libraryData, params }) => {
                       View library assets
                       <ArrowRight size={16} />
                     </Button>{' '}
```
Comment (jdharvey-ibm) [inline services/web-app/pages/libraries/[host]/[org]/[repo]/[library]/[ref]/assets/[asset]/index.js:280]: Should this space still be there if the button next to it is no longer present? I didn't check though so maybe display-wise it doesn't make a difference?

jdharvey-ibm (approved):

services/web-app/pages/libraries/[host]/[org]/[repo]/[library]/[ref]/assets/[asset]/index.js
Comment (alisonjoseph) [reply services/web-app/pages/libraries/[host]/[org]/[repo]/[library]/[ref]/assets/[asset]/index.js:280]: Doesn't appear to make a difference.

Comment (jdharvey-ibm) [reply services/web-app/pages/libraries/[host]/[org]/[repo]/[library]/[ref]/assets/[asset]/index.js:280]: 👍

francinelucca (approved):

PR #1166, opened 2022-08-22
fix(web-app): render remotely loaded SVGs

### LINKED ISSUE #767: SVGs don't work when loading remote MDX pages
Test page with svgs

```
 const mdxSource = await getRemoteMdxData(
    {
      host: 'github.com',
      org: 'carbon-design-system',
      repo: 'carbon-website',
      library: 'carbon-website',
      ref: 'main'
    },
    '/src/pages/guidelines/2x-grid/overview.mdx'
  )
```
Screen Shot 2022-05-23 at 10 52 57 AM

francinelucca:
Closes #767

#### Testing / reviewing

Pages with svgs should now render, remotely loaded images should continue to work

http://localhost:3000/elements/color/implementation
http://localhost:3000/elements/2x-grid/overview


services/web-app/components/markdown/image.js
```diff
@@ -24,11 +24,20 @@ const getImageProps = (props) => {
     props.layout = 'responsive'
   }
 
+  // do not uptimize if is svg file
```
alisonjoseph, line 27:
```suggestion
  // do not optimize if is svg file
```

alisonjoseph (approved):
🔥🔥🔥

services/web-app/components/markdown/image.js
```diff
@@ -24,11 +24,20 @@ const getImageProps = (props) => {
     props.layout = 'responsive'
   }
 
+  // do not optimize if is svg file
+  if (src.split('/').pop().includes('.svg')) {
```
jdharvey-ibm, line 28:
```suggestion
  if (src.split('/').pop().endsWith('.svg')) {
```

jdharvey-ibm (changes requested):

services/web-app/components/markdown/image.js
francinelucca: will this work? urls end with `?raw=true`

jdharvey-ibm: If they end with a query string, then we should parse them via `new URL()` and check the path on that.

francinelucca: did that for absolute urls, since these can be relative as well

services/web-app/components/markdown/image.js
```diff
@@ -24,8 +26,11 @@ const getImageProps = (props) => {
     props.layout = 'responsive'
   }
 
-  // do not optimize if is svg file
-  if (src.split('/').pop().includes('.svg')) {
+  // do not optimize if is svg file (absolute or relative)
+  if (
+    src.endsWith('.svg') ||
+    (isAbsoluteUrl(src) && new URL(src).pathname.split('/').pop().endsWith('.svg'))
+  ) {
```
jdharvey-ibm, line 33: I know this is tempting for brevity, but we should not instantiate objects inside of a conditional expressions. The new url should be defined above instead.

services/web-app/utils/url.js
```diff
@@ -29,3 +29,13 @@ export const urlsMatch = (url1, url2, matchChunks = 0) => {
...
+
+/**
+ * Determines if a url string is absolute
+ *
+ * @param {string} url url string to evaluate
+ * @returns {boolean} whether the url is absolute or not
+ */
+export const isAbsoluteUrl = (url) => {
+  return url.startsWith('http://') || url.startsWith('https://')
+}
```
jdharvey-ibm, line 41: Though somewhat rare, there are other protocols that could be used besides http and https. It would be more comprehensive to try to construct a URL from the given input (`new URL(...)`) and catch an exception indicating an invalid absolute url. Something like:

```ts
let absoluteUrl

try {
  absoluteUrl = new URL(url)
} catch {}

return !!absoluteUrl
```

This might mean that you'd want to move this logic back up to the image.js file (assuming this utility function isn't/won't be used anywhere else).

jdharvey-ibm (changes requested):

services/web-app/utils/url.js
francinelucca: found a similar function in `@utils/string` and using that instead now

services/web-app/components/markdown/image.js
francinelucca: done

services/web-app/components/markdown/image.js
```diff
@@ -27,10 +27,8 @@ const getImageProps = (props) => {
   }
 
   // do not optimize if is svg file (absolute or relative)
-  if (
-    src.endsWith('.svg') ||
-    (isAbsoluteUrl(src) && new URL(src).pathname.split('/').pop().endsWith('.svg'))
-  ) {
+  const url = isAbsoluteUrl(src) ? new URL(src).pathname.split('/').pop().endsWith('.svg') : null
+  if (src.endsWith('.svg') || !!url) {
```
jdharvey-ibm, line 31: For the cases in this PR where `isAbsoluteUrl` is being used and, if true, another URL object is created, it's actually doing more harm than good. Rather than double-instantiating the same URL, this is one of those cases where it's actually a better idea to inline this logic as opposed to using a utility function.

Right now, the logic basically does:
`if new URL(src) doesn't throw and new URL(src).pathname.split....., then ....`

Instead of this, an alternative solution would be to put the instantiation/try/catch right here in this function

```ts
let url
try {
  url = new URL(src)
} catch {}

if (src.endsWith('.svg') || url?.pathname.endsWith('.svg')) {
  ...
}
```

or, the utility function could return the instantiated object as opposed to throwing it away:
```ts
// previously isAbsoluteUrl()
function createUrl(src) {
  let result: URL | undefined
  try {
    result = new URL(src)
  } catch {}

  return result
}

// (later)
const url = createUrl(src)

if (url && someOtherCondition) {
  ...
}
```

jdharvey-ibm (changes requested):

services/web-app/components/markdown/image.js
francinelucca: done!

services/web-app/components/markdown/image.js
```diff
@@ -27,8 +27,8 @@ const getImageProps = (props) => {
   }
 
   // do not optimize if is svg file (absolute or relative)
-  const url = isAbsoluteUrl(src) ? new URL(src).pathname.split('/').pop().endsWith('.svg') : null
-  if (src.endsWith('.svg') || !!url) {
+  const url = createUrl(src)
+  if (src.endsWith('.svg') || (url && url.pathname.split('/').pop().endsWith('.svg'))) {
```
jdharvey-ibm, line 31:
```suggestion
  if (src.endsWith('.svg') || (url && url.pathname.endsWith('.svg'))) {
```

services/web-app/pages/libraries/[host]/[org]/[repo]/[library]/[ref]/pages/[...page].js
```diff
@@ -86,8 +86,8 @@ export const getStaticProps = async ({ params }) => {
 
   let { host, org, repo, ref } = params
 
-  if (isAbsoluteUrl(pageSrc)) {
-    const url = new URL(pageSrc)
+  const url = createUrl(pageSrc)
+  if (url) {
```
Comment (jdharvey-ibm) [inline services/web-app/pages/libraries/[host]/[org]/[repo]/[library]/[ref]/pages/[...page].js:90]: 🔥 👍

services/web-app/utils/string.js
```diff
@@ -65,18 +65,15 @@ export const getUrlWithProtocol = (url) => {
 }
 
 /**
- * Determines if a given string is a valid http url
- * @param {string} urlString
- * @returns {boolean} true if valid url, false otherwise
+ * Converts string to url if valid
+ * @param {string} src
+ * @returns {URL | undefined}}
```
jdharvey-ibm, line 70:
```suggestion
 * @returns {URL | undefined}
```

jdharvey-ibm (approved):

PR #1169, opened 2022-08-23
feat: add additional demo link type other

### LINKED ISSUE #581: Add additional demo link types to the schema
When indexing Carbon Components (vanilla JS), @andreancardona discovered that the demo site was not Storybook but instead a custom demo site: https://the-carbon-components.netlify.app/?nav=accordion

The `demoLinks` schema only specifies `codesandbox`, `github`, and `storybook` as valid `type` values. https://github.com/carbon-design-system/carbon-platform/blob/main/docs/resource-schemas.md#shared-schemas

The schema defines types, so we can have resource cards with an icon per type of demo site. We expect most all libraries to have Storybook. For now, we went with this when indexing Carbon Components:

```yml
demoLinks:
  - type: storybook
    name: Demo
    action: link
    url: https://the-carbon-components.netlify.app/?nav=accordion
```

so the YML file was valid, and so the web app will just say "Demo" in the link.

We either need to make `type` optional, or add a `other` valid value.

**Design**

- [ ] @kingtraceyj to specify what icon we'd use if `other` type or if type is optional and not specified

**Dev**

- [x] Update schema
- [ ] ~Update our Carbon Vanilla indexing so the demoLinks don't say "storybook"~

alisonjoseph:
closes #581

#### Changelog

Add type other to demoLinks schema
Add codereference icon to mdx icon component


francinelucca (approved):
✨

mattrosno (approved):
I didn't test this, but the docs additions look good. I've made a note that after this is merged, we can update demo links that are indexed as `storybook`, but aren't actually Storybook.

PR #1173, opened 2022-08-24
feat(web-app): designKit inheritance

### LINKED ISSUE #1016: Add support for library `designKit` inheritance
Also, https://github.com/carbon-design-system/carbon-platform/pull/831 specifies that a library's `designKits` is inheritable, so a list of design kits can be specified in Carbon Styles library and not every library that depends on it, for example.

francinelucca:
Closes #1016

Libraries now inherit `designKits` array from parent library if `designKits` isn't specified in child library file

#### Changelog

**Changed**

- now using recursive `getLibraryData` call to obtain both inherited assets and designKits in `getLibraryData`

**Removed**

- Some unnecessary cached files

#### Testing / reviewing

Navigate to these 3 routes and confirm `content.designKits` in logged object in browser console matches expected specs:

- http://localhost:3000/libraries/design-kit should have 5 design kits as per defined in https://github.com/francinelucca/library-testing/blob/main/isolated/carbon.yml
- http://localhost:3000/libraries/design-kit-inheritance-2 should have 3 design kits as per defined in https://github.com/francinelucca/library-testing/blob/main/inherit/2/carbon.yml (not inheriting DKs from parent since already defined in file)
- http://localhost:3000/libraries/design-kit-inheritance should have 5 design kits as per defined in parent library (https://github.com/francinelucca/library-testing/blob/main/isolated/carbon.yml) since local library definition does not include `designKits` array https://github.com/francinelucca/library-testing/blob/main/inherit/1/carbon.yml

#### TODO

(On Approval)

- [x] Revert changes in services/web-app/data/libraries.mjs
- [x] Revert changes in services/web-app/pages/libraries/[host]/[org]/[repo]/[library]/[ref]/index.js


alisonjoseph (approved):
Tested and all looks good to me!

mattrosno (commented):
I ran locally and the tests and console logs work as described.

It got me thinking, should `design-kit-inheritance-2` merge or override the inherited design kits? For simplicity, overriding makes most sense, like how this PR handles it. If the base library has design kits that are incompatible with the library that inherits base, then overriding design kits is the only way to handle it and give `design-kit-inheritance-2` full control over what is rendered in its design kits page. Also, if we did a merge, there could be conflicts because it'd be merging objects that could have the same keys, so even the merge wouldn't be clean.

After talking out loud there, I agree with how this is working.

services/web-app/lib/github.js
```diff
@@ -806,6 +789,42 @@ const getLibraryAssets = async (params = {}) => {
   return assets
 }
 
+/**
+ * Validates and filters an array of assets,
+ * returns a modified array containing only valid assets for a library.
+ * @param {Promise<import('../typedefs').Library>} library
```
mattrosno, line 795: Should this be a promise?
```suggestion
 * @param {import('../typedefs').Library} library
```

services/web-app/lib/github.js
```diff
@@ -806,6 +789,42 @@ const getLibraryAssets = async (params = {}) => {
...
+      return isValidAsset && getSlug(asset.content) === library.params.asset
+    }
+    return isValidAsset
+  })
+}
+
+/**
+ * Adds inherited data to library.
+ * @param {import('../typedefs').Library} library
+ * @returns {Promise<void>}
```
mattrosno, line 810: Does this return anything or just modify the param?

mattrosno: Remove the `@returns`?

services/web-app/lib/github.js
francinelucca: true! thankss

services/web-app/lib/github.js
francinelucca: it does return a promise that resolves to undefined, we are awaiting the function when called, I think the @returns notation of an empty promise makes sense here. Double check me on that @jdharvey-ibm ?

mattrosno (approved):

services/web-app/lib/github.js
jdharvey-ibm: image

VS Code agrees! The docs look a little funny not having any descriptive text after the return value though. Maybe just add `@returns {Promise<void>} A promise that resolves to void.`?

francinelucca: Done!

PR #1180, opened 2022-08-25
fix(web-app): build mdx pages dynamically

### LINKED ISSUE #944: Convert library sidenav pages to ISR
Currently generating library sidenav pages at build time, let's switch to ISR when remote MDX epic is complete so that we can have library versioning work

francinelucca:
Closes #944

#### Changelog

**Changed**

- fallback 'blocking' on getStaticPaths to generate page dynamically
- 
#### Testing / reviewing

Test you're still able to render remote pages and lib pages (react tutorial)

### DIFF
--- a/services/web-app/pages/[...remote-mdx-path].js
+++ b/services/web-app/pages/[...remote-mdx-path].js
@@ -80,19 +80,9 @@ export const getStaticProps = async ({ params }) => {
 }
 
 export const getStaticPaths = async () => {
-  const allowedPaths = Object.keys(remotePages).map((remotePage) => {
-    return {
-      params: {
-        'remote-mdx-path': remotePage.split('/')
-      }
-    }
-  })
-
   return {
-    paths: allowedPaths,
-    // returning 404 if page wasn't generated at build time
-    // to prevent remote mdx dynamic loading for now
-    fallback: false
+    paths: [],
+    fallback: 'blocking'
   }
 }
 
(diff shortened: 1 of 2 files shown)


alisonjoseph (approved):
LGTM!

Screen Shot 2022-08-25 at 3 39 19 PM

jdharvey-ibm (changes requested):
Let's chat through the implementation of this together. The core carbon website pages we probably still want to statically build at built time, but then turn on ISR for them so they can change over time without needing to redeploy the app.

The libraries can be statically built on-demand at runtime as you have them below, but we also want to add in the ISR aspect to those as well.

francinelucca:
> Let's chat through the implementation of this together. The core carbon website pages we probably still want to statically build at built time, but then turn on ISR for them so they can change over time without needing to redeploy the app.
> 
> The libraries can be statically built on-demand at runtime as you have them below, but we also want to add in the ISR aspect to those as well.

modified so we're returning allowedPaths in remote-mdx-page so those will get statically generated at built time and added revalidation every hour, do we want to use another interval?

Also added revalidation for library pages (these will not be generated at build time)

jdharvey-ibm:
@francinelucca Every hour sounds good to me!

andreancardona (approved):
tested and looks good 🔥

jdharvey-ibm (approved):

PR #1181, opened 2022-08-25
fix(web-app): filter search popover mobile scroll styles

### LINKED ISSUE #462: Popover needs set max height

alisonjoseph:
closes #462

#### Testing / reviewing

Test filter scroll at mobile view (will need to make height short enough to require a scroll)

http://localhost:3000/assets/components


services/web-app/components/catalog-multislect-filter/catalog-multiselect-filter.module.scss
```diff
@@ -102,9 +102,16 @@
 .content {
   width: calc(100vw - var(--scrollbar-width));
   max-width: none;
+  // height of sticky header + search = 96px
+  height: calc(100vh - convert.rem(96px));
```
francinelucca, line 106: the last section get a little bit cut off for me when scrolled all the way, maybe we should reduce the height a smaller amount? 
image

francinelucca (changes requested):

services/web-app/components/catalog-multislect-filter/catalog-multiselect-filter.module.scss
alisonjoseph: Ah ok, I think the issue is the extra space above the search when you open the filter. Looking into this.

francinelucca (approved):
🚀

mattrosno:
I removed this PR from the sprint, because its associated issue belongs to Sprint 23.

andreancardona:
Going to open an issue for this - but scroll looks good!

Screen Shot 2022-08-29 at 4 01 18 PM

andreancardona (approved):

PR #1188, opened 2022-08-29
fix: bugs found while indexing

mattrosno:
#### Changelog

**New**

- N/A

**Changed**

- Per our asset schema docs (https://next.carbondesignsystem.com/contributing/schema#asset-keys), `tags` is optional. Patterns don't have tags (yet). This PR updates the schema package to allow 0 tags on an asset and not discard it as invalid.
- To accommodate that change, the web app now sets default tags to an empty array, and checks for an empty array when merging inheritable properties.
- I added an en-dash to the Tags column in the assets table, when the asset has no tags. (E.g. https://next.carbondesignsystem.com/libraries/carbon-styles/latest/assets, after the new patterns get merged in.)
- The default framework per the asset schema docs is `design-only`, but we weren't applying that default when ingesting indexed data. This PR fixes that.
- I updated the space between the framework icon and name, to match the space between the status icon and name. It looked like this before:

image

- The asset overview pages weren't showing demo links, probably from a copy/paste error? I fixed that.
- The asset overview page other frameworks list was using framework keys and capitalizing the first character instead of using the framework display names. That resulted in seeing `Design-Only`. I fixed that.

**Removed**

- N/A

#### Testing / reviewing

Main thing to test is that the inherited asset properties still work. Everything else is just visual.


mattrosno:
I got code-smelled by SonarLint.

image

@jdharvey-ibm @francinelucca me adding the default `design-only` and `[]` I guess added too much complexity... is there an easy fix besides moving those out to new functions to set defaults?

francinelucca:
> I got code-smelled by SonarLint.
> 
> image
> 
> @jdharvey-ibm @francinelucca me adding the default `design-only` and `[]` I guess added too much complexity... is there an easy fix besides moving those out to new functions to set defaults?

Don't know if this will work but I'd try doing `get(asset, ['framework'], 'design-only')`  instead of `asset?.framework ?? 'design-only'`  (get from lodash). and same for the tags

mattrosno:
@francinelucca `get(asset, 'framework', 'design-only')` worked. The best way to test that is `carbon-styles` library seeing all the `Design only` tags, that aren't in main branch: https://next.carbondesignsystem.com/libraries/carbon-styles/latest/assets

image

services/web-app/components/dashboard/dashboard.module.scss
```diff
@@ -110,7 +110,7 @@
   top: convert.rem(-3px);
   width: 1rem;
   height: 1rem;
-  margin-right: spacing.$spacing-03;
+  margin-right: spacing.$spacing-02;
```
alisonjoseph, line 113: Did you confirm this with design? It doesn't match the redline, but there isn't a specific spacer noted in Figma either. 

https://www.figma.com/file/vtmHnw9ACgnksFg7ZvEsZC/RL_Asset_ComponentDetails?node-id=0%3A1

Screen Shot 2022-08-30 at 12 01 43 PM
Screen Shot 2022-08-30 at 12 01 57 PM

alisonjoseph: @kingtraceyj

mattrosno: @kingtraceyj @jeanservaas 

8px spacing:
image

4px spacing:
image

This PR updates both to be 4px spacing, so the framework icon spacing matches the status icon, which was in there first.

Is that okay?

mattrosno: This PR standardizes on using `$spacing-02` between icon and text in these two places.

alisonjoseph (approved):
Everything looks like its displaying correctly (other than just double checking if those icons should be 4px or 8px spacing). 👍

kingtraceyj:
status should be 4px padding according to the status indicator pattern:
image


if possible we can use the `CircleSolid` icon for a true icon. it also has padding built into it.

Since we going use the 4-x padding for the status we should also use it for the framework icon.

mattrosno:
@kingtraceyj we're using CircleFill. See: https://github.com/carbon-design-system/carbon-platform/blob/main/services/web-app/data/status.js

We're using the official "status" icon for this:

image

Where CircleSolid is in the formatting section:
image

The SVG looks to have the proper internal padding:
image

mattrosno:
Thanks @kingtraceyj. I resolved @alisonjoseph, since 4px spacer is good per @kingtraceyj.

jdharvey-ibm:
> is there an easy fix besides moving those out to new functions to set defaults?

@mattrosno Moving some chunks of functionality out into separate functions would 100% be the best approach. Is there something preventing that from being a viable solution? The lodash function really doesn't improve the situation at all. If anything, the complexity is now even higher because the lodash function is a bit esoteric.

mattrosno:
> Moving some chunks of functionality out into separate functions would 100% be the best approach. Is there something preventing that from being a viable solution?

@jdharvey-ibm just time? I can certainly refactor into new functions if necessary. We use the `get` lodash functions in ~6 web-app files, so we have precedent there on using it (get a value in an object and specify a default if there are any errors getting that value)?

mattrosno:
@jdharvey-ibm here's my refactor commit to not use Lodash `get` and not add SonalLint complexity: https://github.com/carbon-design-system/carbon-platform/pull/1188/commits/4050b862d697f50366baf87832b6833360452066

francinelucca (approved):
LGTM

jharvey10 (approved):
Refactor looks good to me!

PR #1193, opened 2022-08-30
Update team names

### LINKED ISSUE #963: Update maintaining teams
After https://github.com/carbon-design-system/carbon-platform/issues/951 is completed, we should consider these updates.

We currently have 7 maintaining teams specified here: https://github.com/carbon-design-system/carbon-platform/blob/main/services/web-app/data/teams.js#L29

**Update**

- [x] `Carbon` to `Carbon System Squad`
- [x] `IBM.com` to `Carbon Dotcom Squad`

**Add**

- [x] `IBM Accessibility` for the two design kits that team maintains (we'd also need a icon and pictogram)

**Justification**

The five indexed IBM.com libraries are maintained by the Carbon team but it just reads "IBM.com" so it's not extremely clear what that means. Hill 1A is about having confidence in using assets by knowing who maintains. Being more specific here should help instill confidence. Plus, when you view library details, we'll list out maintainers on a per-name basis to provide even more specificity.

E.g.:
image

mattrosno:
Closes #963 

#### Changelog

**New**

- IBM Accessibility pictogram as custom SVG until this comes from Brand (I tested this pictogram locally and it works. We won't be showing this anywhere in v1 since we don't have design kit detail pages.)

**Changed**

- Carbon team names per issue

**Removed**

- N/A

#### Testing / reviewing

You see updated maintaining team names for Carbon core-maintained assets.


francinelucca (approved):
Confirmed can see new team names.

@mattrosno won't the accessibility icon show up in the design catalog page? or will we not have ibm accessibility DKs indexed for v1?

alisonjoseph (approved):

mattrosno:
@francinelucca the team icon shows up in the catalogs, but not the pictograms.

PR #1203, opened 2022-08-31
fix(web-app): cache schema resolved references

### LINKED ISSUE #971: Ensure resolved references are cached
The `/lib/github.js` functions initiate GitHub API requests and directly cache the responses.

Because `@apidevtools/json-schema-ref-parser` makes network requests, we want to wait to cache the GitHub API responses until after all `$ref`s have been dereferenced.

To accomplish this, one approach would be to add logic to `_getResponse()` in `/lib/file-cache.js`, so if the request is using the GitHub contents API, if requesting a `carbon.yml` file, dereference the content in that function and return the dereferenced content so the modified response gets cached.

francinelucca:
Closes #971

made this very generic as dereferencing is very object-structure dependant and there is potential plans to expand refs across all our schemas in future releases

#### Changelog

**New**

- `getDereferencedObjectResponse` in `file-cache.js` to handle caching objects after they've been dereferenced given a key
- new cached files for dereferenced libs

**Changed**

- now using `getDereferencedObjectResponse` in `github.js` when getting library data

#### Testing / reviewing

you can add an entry into `libraries.mjs` and verify the cached entries that are generated. for this one there should be an entry with the key = `"key": "library=github.com/francinelucca/library-testing&path=/&ref=main-dereferenced"` that should contain the library object with the design kits array populated (without refs)
```
  'design-kit': {
    host: 'github.com',
    org: 'francinelucca',
    repo: 'library-testing',
    path: '/'
  }
```


francinelucca:
@mattrosno  'd love to get your thoughts  on this

-  made just a generic deferencing function in the file-cache that takes a callback because noticed the logic to dereference is very specific to the library object and since we probably want to use refs more widely across the schema in the future it'd make more sense to let the caller handle that. 
- Also chose to still cache the original github response just in case there's some duplicate requests from other libraries (could see this happening if multiple libraries reference a design kit url for example)

mattrosno:
@francinelucca I'll review first thing tm!

mattrosno (commented):
I see `library=github.com/francinelucca/library-testing&path=/&ref=main-dereferenced` in the cache. 👍 

image

services/web-app/lib/github.js
```diff
@@ -647,22 +647,26 @@ export const getLibraryData = async (params = {}) => {
...
+      repo,
+      path: removeLeadingSlash(`${libPath}/carbon.yml`),
+      ref
     })
   } catch (err) {
     return null
   }
 
   let content
+  const dereferenceKey = `library=${host}/${org}/${repo}&path=${libPath}&ref=${ref}-dereferenced`
```
mattrosno, line 664: `file-cache.js` has a private `slugifyRequest()` that generates cache keys. Would it be good to use that function to create the dereference key?

mattrosno: Maybe not. I see those keys look like:

`github.com GET /repos/francinelucca/library-testing/contents/carbon.yml ref=main`

Instead of:

`library=github.com/francinelucca/library-testing&path=/&ref=main-dereferenced`

Maybe match that format with:

`github.com GET DEREFERENCED /repos/francinelucca/library-testing/contents/carbon.yml ref=main`

?

Not a big deal. I can't think of a scenario when your key wouldn't be unique. If you want to reuse `slugifyRequest()`, you could just pass in if it's a dereferenced request.

mattrosno (commented):
The original cached response is base 64-encoded content:
image

The dereferenced data that gets cached is an object, not base 64-encoded. I think that's okay. Just making sure?
image

mattrosno (commented):
Using

```
  'design-kit': {
    host: 'github.com',
    org: 'francinelucca',
    repo: 'library-testing',
    path: '/'
  }
```

I see the 5 kits.

image

I committed (but didn't push) all the `/public/data` for a baseline. I restarted dev environment and went to that page again.

Confirmed that I see the 5 kits now, and that there are no newly-cached files to commit.

image

So this looks like its all working for me!

mattrosno (approved):
No strong opinion on the cache key format. Just something to consider.

services/web-app/lib/github.js
francinelucca: using `slugifyRequest` to generate key now!

francinelucca:
@mattrosno now storing in base64 💃

PR #1232, opened 2022-09-08
fix(web-app): add anchor element to tree nav

### LINKED ISSUE #1124: Navigable elements in left-nav are not anchor links
**Description**

I noticed when trying to middle click a left-nav link to open it in a new tab, that the navigation elements in the nav are not anchor links, but are instead `div`s inside of `li`s.

I haven't looked into the code, but I think what might need to happen is that if a tree node has no "items" (i.e. it is a leaf node), it should instead be rendered as an anchor.

alisonjoseph:
closes #1124

Adds anchor link for nav items

#### Testing / reviewing

Test that left nav looks / behaves the same as the live site and that anchor links are rendering for links


francinelucca (changes requested):
LGTM!
QQ, should we make the anchor span the whole space? for example FAQ will only work if I click directly on the text, if I click next to it (but within the tree item) it won't

alisonjoseph:
@francinelucca good call! updated the styling

francinelucca (changes requested):
URL looks good across the item now but I see a height difference when compared to prod 🤔 
image
image

alisonjoseph:
@francinelucca  oops 🤦‍♀️ try now

francinelucca (approved):
Looks great! I think you also fixed #1248

andreancardona (approved):
💯

PR #1263, opened 2022-09-13
fix(web-app): revalidate statically generated pages, add cache for SSG

### LINKED ISSUE #1200: Convert all SSG pages that use remote content to ISR
**Summary**

What problem (or opportunity) are we solving? Why are we solving it? Is there supporting user
conversations or research?

We want content to be updated on production without having to redeploy, therefore let's have pages revalidate every hour.

**Job stories**

What value will this deliver?

When remote content consumed by the application is changed, I want to see the updated versions in the web app without the need of a deployment, so I can read the most up to date content.

**Acceptance criteria**

What requirements must be met to complete this request?

1. getStaticProps pages have a revalidation of one hour across the app if they are reading remote content (whether its our own data graph, github, medium, etc...)) (most are)

francinelucca:
Closes #1200 

Added revalidation to statically generated pages and converted server side generated pages to ISR

#### Testing / reviewing

 > There should be no regressions on the app, everything should look/behave exactly the same


francinelucca:
@jdharvey-ibm does this cache strategy make sense for the server side generated pages? 🤔 or should we leave them as-is? could also convert them to ISR but I don't think we can prevent them from building the first time if we do that and that's a LOT of pages :(.

https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props#caching-with-server-side-rendering-ssr

alisonjoseph:
Could the lint error be caused by the comments inside of the `mdx` files? I think we might need to use `/*` instead of `// `but I'm not 100% sure. I'm not getting lint errors locally. 😅

jdharvey-ibm (approved):

alisonjoseph (approved):

PR #1264, opened 2022-09-13
fix(web-app): fixes undefined maintainer in library

### LINKED ISSUE #1204: Web-App: component catalog, broken filter by "Community"
**Description**

Tell us more about the problem that you're running into. What did you see? What did you expect to
see?

We definitely have assets that are community maintained indexed, yet when we filter by community it's coming up as zero results.

*  Consider doing something like this for libraries to guarantee every lib has a maintainer https://github.com/carbon-design-system/carbon-platform/blob/main/services/web-app/data/design-kits.js#L155
* Given ^, consider refatoring/cleaning up this logic since maintainer will be guaranteed https://github.com/carbon-design-system/carbon-platform/blob/main/services/web-app/components/asset-catalog-item/asset-catalog-item.js#L56 (56-60)

**Steps to reproduce**

1. https://next.carbondesignsystem.com/assets/components?maintainer[]=community, notice no results

francinelucca:
Closes #1204

#### Changelog

**New**

 - add default maintainer to libraries

**Changed**

- cleanup asset maintainer icon logic

**Removed**

- {{ Removed thing }}

#### Testing / reviewing

test maintainer icon tooltip and make sure non-community assets still work as expected

http://localhost:3000/assets/components?maintainer[]=community

### DIFF
--- a/services/web-app/components/asset-catalog-item/asset-catalog-item.js
+++ b/services/web-app/components/asset-catalog-item/asset-catalog-item.js
@@ -5,7 +5,6 @@
  * LICENSE file in the root directory of this source tree.
  */
 import { AspectRatio, Column, Grid } from '@carbon/react'
-import { Events } from '@carbon/react/icons'
 import clsx from 'clsx'
 import Link from 'next/link'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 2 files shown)


andreancardona (approved):
💯

alisonjoseph (approved):
👍👍

PR #1283, opened 2022-09-19
fix(web-app): gracefully handle prismjs crashes

jharvey10:
Fixes the issue with the schemas page crashing in a minified build.

#### Changelog

**New**

- include a prism download from prismjs.com (recommended approach until full esm support is added)
- - highlight code via prism in a useEffect block in the code component

**Changed**

- ‼️ switch to non-wrapping code blocks
  - [x] verify that this behavior is okay and is an improvement

**Removed**

- remove port number from next dev command so it can be manually specified

#### Testing / reviewing

Check visual style of code blocks across a variety of pages to ensure it looks the same as before (except for the new lack of wrapping)


services/web-app/styles/_prism-theme.scss
```diff
@@ -0,0 +1,160 @@
...
+}
+
+.token.namespace {
+  opacity: 0.7;
+}
+
+/* LANGUAGE-SPECIFIC OVERRIDES */
+// .language-css .token.important {
+//   color: unset;
+// }
```
mattrosno, line 160: Did you mean to remove this?

jdharvey-ibm: I wanted to keep it there for future reference in case we ever need to add custom language colors. I can remove it if you'd prefer that.

services/web-app/public/prism.js
```diff
@@ -0,0 +1,8 @@
+/* PrismJS 1.29.0
```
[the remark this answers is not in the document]
jdharvey-ibm: I definitely thought about that and we can certainly add it. I took what their site provided verbatim, assuming that if they wanted their license in the minified version, they would have included it. The URL at the top does link back to their site, which has a paper trail to their license and the mechanism by which we downloaded it.

services/web-app/styles/_prism-theme.scss
mattrosno: Oh, no preference, maybe a comment would help "for future reference, if we need to... here's an example of what we'd need to do...".

services/web-app/public/prism.js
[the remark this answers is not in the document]
mattrosno: Or, we save both prism.js and prism.min.js so we can save the MIT comment, and then only use the minified version in the web app.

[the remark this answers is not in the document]
jdharvey-ibm: 👍 I can add the non-minified version.

services/web-app/components/code/code.module.scss
```diff
@@ -27,22 +26,11 @@
...
-  // override carbon inline style
-  min-height: 24px !important;
-
-  pre {
-    padding-bottom: 0;
-  }
-
-  code {
-    background-color: theme.$layer-01;
-  }
```
alisonjoseph, line 45: I think we need these styles?

alisonjoseph (changes requested):
I'm seeing a lot styling errors. I think it has to do with some of the lines removed 33-45 in code.module.scss file? Also looks like we've lost syntax highlighting on some of the code blocks that had it previously. (was testing this page http://localhost:3000/libraries/carbon-react/latest/pages/tutorial/step-2)

Screen Shot 2022-09-20 at 7 29 46 AM
Screen Shot 2022-09-20 at 7 30 08 AM
Screen Shot 2022-09-20 at 7 30 30 AM
Screen Shot 2022-09-20 at 7 30 51 AM

alisonjoseph:
> ‼️ switch to non-wrapping code blocks 

Verifying this is what we want and a bug on our live site 👍

services/web-app/components/code/code.module.scss
```diff
@@ -9,54 +9,41 @@
...
 @use '@carbon/react/scss/motion' as motion;
-/* stylelint-disable declaration-no-important */
-
-// scoped to div code to override global reset styles
-:global(div) code {
-  padding: 0 spacing.$spacing-03;
-  background-color: theme.$layer-accent;
-  border-radius: 4px;
-  font-size: 0.875em;
-}
```
alisonjoseph, line 20: Missed this before, these styles are styling the base inline code in markdown, we aren't translating them to the inline Carbon code snippet for some reason. Probably should be vs. creating our own styles here, trying to remember why I did it this way. 

Screen Shot 2022-09-20 at 2 06 02 PM

francinelucca (changes requested):
Looks Great!
Only thing is I see a dark background on md breakpoint that I don't see on prod 
image

services/web-app/components/mdx/components.js
```diff
@@ -63,6 +63,9 @@ const components = {
   img: Image,
   table: PageTable,
   pre: Code,
+  code: ({ children }) => (
+    <mdxComponents.CodeSnippet type="inline">{children}</mdxComponents.CodeSnippet>
```
alisonjoseph, line 67: Nit-picky, but should we pull this out into its own component?

alisonjoseph (approved):
Working great 🥳  just one small question/comment

services/web-app/components/mdx/components.js
jdharvey-ibm: I definitely think we should! We could add a `Code` and `InlineCode` component to the mdx-components package. As we keep adding stuff over there, this file will get smaller to reflect that. Would that work? Or would you prefer to address it in this PR first and then move it to the mdx-components package as a second step?

alisonjoseph: I'm fine with doing it when we move the Code component over to the mdx-components package.

francinelucca (approved):
✨

PR #1331, opened 2022-09-20
feat: add isr fallbacks

mattrosno:
Closes #1313

This PR enables ISR fallbacks.

#### Changelog

**New**

- `PageLoading` component with a multiple line skeleton

**Changed**

- `NavPrimary` skeleton when there's no primary nav data
- `NavSecondary` skeleton when there's no secondary nav data
- `PageHeader` loading state
- Fallbacks used in ISR pages:
  - Remote MDX
  - Library
  - Library > Assets
  - Library > Design kits
  - Library > Page
  - Library > Asset
  - Library > Asset > Tab

**Removed**

- N/A

#### Testing / reviewing

See comments below for examples.

PR #1356, opened 2022-09-26
feat: add loading page with skeletons

mattrosno:
- add loading to side navs
- add loading to page header

See https://github.com/carbon-design-system/carbon-platform/pull/1331 description.

This is a new PR to replace 1331 now that the `WithLoading` component is in main branch.


services/web-app/components/with-loading/with-loading.js
```diff
@@ -7,24 +7,22 @@
...
-      </div>
+    <Column sm={4} md={8} lg={8} {...config?.column}>
+      <PageHeader loading {...config?.pageHeader} />
+      <PageLoading />
     </Column>
   </Grid>
 )
 
-const WithLoading = (Component) => {
+const WithLoading = (Component, config = {}) => {
```
jdharvey-ibm, line 22: Since this isn't bound by propTypes, could we add some docs to this HOC and/or a typedef for its parameters?

mattrosno: @jdharvey-ibm want something like this?
image

mattrosno: I just pushed a commit with these changes.

jdharvey-ibm: @mattrosno I'm not sure that `WithLoading` propTypes object has the intended result. 

It's defining the props as a single object being comprised of a `Component` key and a `config` key. This is different than how it's being used, which is a two-parameter function.

Since `WithLoading` is an HOC and not a functional component, I don't believe propTypes can be used on it.

@francinelucca This is minor, but I was looking through the react docs, and it looks like their convention for HOC naming is lower camel case (first letter lowercased). Could we make an issue to rename that component from `WithLoading` to `withLoading`?

mattrosno: @jdharvey-ibm how about this? Manually checking propTypes.

image

mattrosno: Pushed with `withLoading` capitalization change.

jdharvey-ibm: With those changes, are you able to get IDE auto-completion? That's the main thing I was hoping for when asking about adding a typedef.

mattrosno: You get browser console errors if invalid prop types like normal React components. That's what I was after.

alisonjoseph (approved):
loading states look good and appear to be working correctly 👍

services/web-app/components/with-loading/with-loading.js
francinelucca: I feel kind of weird about having `withLoadingPropTypes` that are just types for the second `config` param here, I'd do JsDocs and define a custom type instead 🤔 , is that what you were saying? @jdharvey-ibm

jdharvey-ibm: @francinelucca Yes! Exactly. Just something to help during development so we know how to use the parameters to that HOC.

mattrosno: @jdharvey-ibm do I need to update to JSDoc or is the PropTypes good?

jdharvey-ibm: The PropTypes doesn't address my original comment of documenting the `withLoading` component. I'd prefer if we just document withLoading as a normal JS function, since it's not a normal functional react component. If the config parameter can be documented inline without the use of a typedef, then that's totally fine, but if it can't, we should create a typedef for that object like we have for a lot of the github.js input/result objects.

mattrosno: Pushed new commit with JSDoc comment.

andreancardona (approved):
🎉

jdharvey-ibm (approved):

PR #1359, opened 2022-09-27
fix(web-app): remove top border in library catalog first item

### LINKED ISSUE #1301: Libraries catalog that there is a rule on the top of the very first card. This is not to spec. There should not be a rule there.
This is a visual bug, but I'm noticing in the Libraries catalog that there is a rule on the top of the very first card. This is not to spec. There should not be a rule there.

Page url: https://next.carbondesignsystem.com/libraries

image

_Originally posted by @jeanservaas in https://github.com/carbon-design-system/carbon-platform/discussions/1268#discussioncomment-3684172_

francinelucca:
Closes #1301

We actually steered away from the gh pages route and are now deploying directly to an IBM Cloud Object Storage bucket that servers the files as a static website, which saves us the trouble of having to create a new repository to host the files 🔥 .

We'll do the same thing for pr storybook deployments using a different bucket

#### Testing / reviewing

http://localhost:3000/libraries

### DIFF
--- a/services/web-app/components/library-catalog-item/library-catalog-item.js
+++ b/services/web-app/components/library-catalog-item/library-catalog-item.js
@@ -6,7 +6,9 @@
  */
 import { AspectRatio, Column, Grid } from '@carbon/react'
 import { Events, Scales } from '@carbon/react/icons'
+import clsx from 'clsx'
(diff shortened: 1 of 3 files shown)


alisonjoseph (approved):
LGTM

Issue #1361, opened 2022-09-27
Nav active state is lost when opening another nav section


alisonjoseph:
When on a page and opening another section the active state should stay active for the page you are on.

https://user-images.githubusercontent.com/2753488/192604864-a3034d22-4276-4d80-b8a5-d14a3b27c731.mov

alisonjoseph:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-Development-624c3fc8b7972e001130fef7/issues/carbon-design-system/carbon-platform/1361?planning-poker) with Zenhub @andreancardona @francinelucca @jdharvey-ibm

PR #1362, opened 2022-09-27
fix(web-app): remove nav active state from section headings

### LINKED ISSUE #1287: Nav item shouldn't be "Active" when you open it but don't navigate to the page
**Bug🐛:** Nav item shouldn't be "Active" when you open it but don't navigate to the page
**Page:** https://next.carbondesignsystem.com/
**Suggestion:** I'm on the homepage, opened the "Designing" section and it set the active styles, it should only open the section, and set active when you are actually on that page. The headings that aren't pages shouldn't ever have the active state with the blue border to the left. See current Carbon website nav for example. 
Screen Shot 2022-09-16 at 9 41 49 AM

_Originally posted by @alisonjoseph in https://github.com/carbon-design-system/carbon-platform/discussions/1268#discussioncomment-3663385_

alisonjoseph:
Closes #1287

#### Changelog

**Changed**

- treeview nav styles

#### Testing / reviewing

Test navigation, section headings should no longer take on active class styles (no background or blue left border)

### DIFF
--- a/services/web-app/components/nav-tree/nav-tree.module.scss
+++ b/services/web-app/components/nav-tree/nav-tree.module.scss
@@ -113,6 +113,10 @@
     color: theme.$text-primary;
   }
 
+  :global(.cds--tree-node[aria-expanded='true'] > .cds--tree-node__label) {
+    @include type.type-style('heading-01');
+  }
+
   .section-heading {
     padding-bottom: spacing.$spacing-03;
     border-bottom: 1px solid theme.$layer-accent;
--- a/services/web-app/components/nav-tree/nav-tree.module.scss
+++ b/services/web-app/components/nav-tree/nav-tree.module.scss
@@ -14,15 +14,7 @@
 .container {
   margin-bottom: spacing.$spacing-10;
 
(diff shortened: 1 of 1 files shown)


francinelucca (commented):
is this PR supposed to leave the activeItem active or just remove the active state from the sections? I no longer see the active state on the sections but the active item's active state is also removed 🤔 
image

alisonjoseph:
@francinelucca I'm still seeing the active state? It goes away when you open a section, but it does that currently also on the live site.

francinelucca:
@alisonjoseph yeah I was just asking since now the active item won't be in the selected item, should the active item stay in the current page? and if so, is that in the scope of this ticket or is there another?

francinelucca:
> @alisonjoseph yeah I was just asking since now the active item won't be in the selected item, should the active item stay in the current page? and if so, is that in the scope of this ticket or is there another?

Tracked in #1361 , this is good to go. Approving

francinelucca (approved):

andreancardona (approved):
tested and looks good!

PR #1368, opened 2022-09-28
fix: remove nav skeletons

mattrosno:
Removes primary and secondary nav skeletons while data is loading, because that becomes distracting as the entire web-app re-hydrates between page loads now.

#### Changelog

**New**

- NA

**Changed**

- NA

**Removed**

- Primary and secondary nav loading skeletons

#### Testing / reviewing

Navigate among about pages, and pages in a library, and make sure that side navs remain white while loading.


francinelucca (approved):

alisonjoseph (approved):
LGTM

PR #1407, opened 2022-10-05
feat(web-app): add global content wrapper for consistent spacing

### LINKED ISSUE #1215: Investigate global wrapper component for page padding-top
Looks like we lost the correct padding at the top of the page when switching from mdx to js page. Wondering if we should move this style to a global wrapper, although may add incorrect spacing to other pages if we do that. Can probably look at that in a different issue unless its an easy fix.

Screen Shot 2022-09-01 at 9 56 19 AM

Screen Shot 2022-09-01 at 9 58 20 AM

Everything else is running / working great!

_Originally posted by @alisonjoseph in https://github.com/carbon-design-system/carbon-platform/pull/1205#pullrequestreview-1093558542_

alisonjoseph:
closes #1215

New `ContentWrapper` component.
  - The intention is for any page content, mdx or otherwise to be wrapped in this component for consistent top spacing. 

This component was NOT added to any of the catalog pages for now, those pages have a different top spacing because the search bar is the first element. 

#### Testing / reviewing
Check mdx pages, js page and remote mdx pages, pages with/without tabs etc. 

http://localhost:3000/elements/icons/library
http://localhost:3000/libraries/carbon-react/latest/assets/accordion
http://localhost:3000/libraries/carbon-react/latest/assets/accordion/style
http://localhost:3000/libraries/carbon-react/latest/assets/accordion/usage
http://localhost:3000/libraries/carbon-react/latest/assets/accordion/accessibility
http://localhost:3000/about-carbon/releases
http://localhost:3000/about-carbon/articles
http://localhost:3000/elements/color/usage
http://localhost:3000/libraries/carbon-components-angular/latest
http://localhost:3000/libraries/carbon-components-angular/latest/design-kits
http://localhost:3000/libraries/carbon-components-angular/latest/pages/tutorial/step-2

### DIFF
--- a/services/web-app/components/asset-details/asset-details.js
+++ b/services/web-app/components/asset-details/asset-details.js
@@ -13,6 +13,7 @@ import { useRouter } from 'next/router'
 import { MDXRemote } from 'next-mdx-remote'
 import PropTypes from 'prop-types'
 
+import ContentWrapper from '@/components/content-wrapper'
 import { Dashboard, DashboardItem } from '@/components/dashboard'
 import dashboardStyles from '@/components/dashboard/dashboard.module.scss'
 import DemoLinks from '@/components/demo-links'
@@ -62,7 +63,7 @@ const AssetDetails = ({ library, asset }) => {
   const router = useRouter()
 
   return (
-    <>
+    <ContentWrapper>
       <section id="dashboard">
         <Dashboard className={styles.dashboard}>
           <Column className={dashboardStyles.column} sm={4}>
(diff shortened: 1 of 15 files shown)


jdharvey-ibm (approved):
I didn't test it locally, but the code changes look good to me 👍

francinelucca (changes requested):
This is GREAT 👯‍♀️ , just one thing: 
Looks like the first item rule is messing with the dashboard item spacing:
image
image
---

alisonjoseph:
@francinelucca good catch! I updated it to only zero out the margin-top for first elements, this way we can preserve any custom spacing needed with padding. Currently all the core mdx components use margin-top to space things out, so this should work.

francinelucca (approved):
LGTM!

PR #1417, opened 2022-10-05
feat(web-app): remove team photos

jdharvey-ibm:
Since we no longer have a team page on the website, these files are not used anywhere. If we need them back in the future, we can go back to the commit before this one and retrieve them.


alisonjoseph (approved):
Makes sense to me 👍

francinelucca (approved):
💪

PR #1418, opened 2022-10-05
feat(web-app): have MdxPage take a source string instead of children

jdharvey-ibm:
- This reduces the number of places where MDXRemote needs to be used
- MdxPage now handles this for "remote-mdx-path" and "page"
- Fix docs in mdx utils


jdharvey-ibm:
This won't actually work until we deprecate the `withMDX` loader from the nextjs config.

PR #1464, opened 2022-10-28
fix(web-app): remove tab strops from hidden sidenav

### LINKED ISSUE #1296: Keyboard navigation is accessing hidden links
**Bug🐛:** keyboard navigation is accessing hidden links 
When using the tab key to navigate from the UI shell to the side panel, I am seeing the keyboard access `/developing` `/design-kits` `/libraries` before focus is on "back to" button. 
Then the focus state is sometimes not styled correctly when using tab key vs up and down arrows (see the focus of Assets and Design kits
platform--navigation--Sep-19-2022 16-30-20

**Page:** Noticed on the [IBM.com React library](https://next.carbondesignsystem.com/libraries/ibmdotcom-react) but should check on all pages that utilize the "back to" button
**Browser:** Chrome
**Suggestion:** disable the first treeview when another one is loaded on top so you don’t tab through invisible links

_Originally posted by @oliviaflory in https://github.com/carbon-design-system/carbon-platform/discussions/1268#discussioncomment-3684605_

francinelucca:
Closes #1296

#### Changelog

**Changed**

- Disable treenodes and add tabIndex -1 when sidenav is hidden

#### Testing / reviewing

Test both sidenavs


services/web-app/components/nav-tree/nav-tree.js
```diff
@@ -55,6 +55,15 @@ const NavTree = ({ activeItem, items = [], label }) => {
...
 
+  useEffect(() => {
+    requestAnimationFrame(() => {
+      // Remove tabstop from headings
+      document
+        .querySelectorAll('[class*=section-group]')
+        .forEach((heading) => heading.setAttribute('tabindex', '-1'))
+    })
+  }, [])
+
```
francinelucca, line 66: The `h2`s inside the treeview where showing up with `tabIndex=0` and I couldn't figure out why, my theory is it's the treeview doing it behind the scenes and this is the only way I could figure out to override them to remove that hidden tabStop, open to any suggestions though...

alisonjoseph (approved):
working correctly for me

andreancardona:
Works for me too but happy to look at this closer @francinelucca if you think it's worth it? let me know :)

services/web-app/components/nav-tree/nav-tree.js
francinelucca: leaving this as-is hehe

andreancardona (approved):
🔥

PR #1582, opened 2023-02-01
deps(mdx-sanitizer): pull Node import from unist

francinelucca:


andreancardona (approved):

mattrosno (approved):
I see we are doing a `Node` only import here:

https://github.com/carbon-design-system/carbon-platform/blob/2791e81c2edb2f2515c49779dffba7b1137c70f6/packages/mdx-sanitizer/src/main/interfaces.ts#L8

I didn't verify this fix, but looks to be a valid import.

PR #1601, opened 2023-02-09
fix(web-app): handle noIndex at asset level,inherit default from library

### LINKED ISSUE #1593: Discrepancy with the tags
**Description**

Discrepancy with the tags - I've so far only noticed it in `Patterns with Angular` directly from the search page - ex: if you filter by `Angular` you won't see the bug 

1. Nav to this link: https://next.carbondesignsystem.com/assets/patterns
2. Scroll down to a pattern that has an `Angular` tag - it shows as if this pattern is available in  other frameworks but then navigating to `Dialogs` (for example)  it only has 1 framework: https://next.carbondesignsystem.com/libraries/carbon-components-angular/latest/assets/dialogs

Screenshot 2023-02-08 at 13 12 34

francinelucca:
Closes #1593 

Refine `noIndex` strategy:
- if an asset has noIndex and its library is not noIndex:true, then an individual asset can be hidden
- if a library has noIndex:true, and assets don't specify noIndex... they follow their parent and are also hidden
- if a library has noIndex:true, but an asset explicitly states noIndex:false, then the asset would show up in the assets catalog but the library would not show up in the libraries catalog

#### Changelog

**Changed**

- Asset will now inherit `noIndex` property from library but can override at the asset level

**Removed**

- mergeAssetContentDefaults function, centralized instead in `addAssetDefaults`

#### Testing / reviewing

Bug defined in #1593  should be fixed, the catalogs should otherwise be displaying the same content, compare to next.carbondesignsystem.com to ensure no content regressions with new `noIndex` logic


francinelucca:
Need to merge https://github.com/carbon-design-system/carbon/pull/13127 in order for this to be properly testable

andreancardona (approved):
looks good to me  - irrelevant - but do we want to think about renaming the `design-only` tag? Doesn't seem to be a good indicator for what it is 👀

andreancardona (approved):

francinelucca:
> looks good to me - irrelevant - but do we want to think about renaming the `design-only` tag? Doesn't seem to be a good indicator for what it is 👀

Created an issue https://github.com/carbon-design-system/carbon-platform/issues/1604

alisonjoseph (approved):

PR #1670, opened 2023-04-21
fix: tree view style overrides

mattrosno:
Closes #1668 

Upstream component had styling changes so we have to modify our styling overrides

#### Changelog

**New**

- N/A

**Changed**

- Wrapped tree view in the g10 theme because that looks like a better starting point for overrides
- Tree view style overrides

**Removed**

- N/A

#### Testing / reviewing

Run locally and go to pages and subpages, desktop and mobile, and ensure that the side nav looks like the production site.

PR #1672, opened 2023-04-21
feat: hide guidance pages

mattrosno:
While Carbon Platform's web app remains a public beta on the next subdomain, we're removing the guidance pages so:

* We don't have to keep content in sync between carbon-website and carbon-platform repositories
* When content is not in sync, this prevents users from referencing the wrong guidance
* It simplifies the RMDX effort by reducing the number of remote MDX pages currently getting rendered

https://github.com/carbon-design-system/carbon-platform/pull/1670 should get merged first.

Instead of deleting pages, this PR simply removes pages from the side nav. This allows pages to be referenced when updating the Gatsby website. The entire site has search engine crawling disallowed while it's on a subdomain ([robots.txt](https://next.carbondesignsystem.com/robots.txt)), so there's no harm in keeping the routes active.

Pages and/or content that should be ported to the Gatsby website. We might want to do this before merging this PR.
- [ ] https://next.carbondesignsystem.com/about-carbon/how-carbon-works
- [ ] https://next.carbondesignsystem.com/about-carbon/community-sites

It was requested to copy this page to the Gatsby website. It uses a `FilterableDesignKitTable` component that's only in Carbon Platform, which uses indexed schema data. Because it's not a simple markdown / MDX page, the level of effort is greater than low.

* https://next.carbondesignsystem.com/designing/get-started

#### Changelog

**New**

- N/A

**Changed**

- The "return to classic website" banner now shows on all routes, not just the home page, to always guide users back to Carbon's default site
- Hid the home page feature card for how carbon works, the section to start designing and start developing, and the section that links to the roadmap because those pages have been hidden

**Removed**

- Hid all side nav items but the Contributing > Schema page and the catalogs
- The top pages for About Carbon, Designing, Developing, Contributing, Migrating, Elements, and Guidelines pages can all be found in the Gatsby website
- The Data visualization collection was also removed (hidden) because all of the content can be found in the Gatsby site, and because you can view the components in the catalog filtered by `Data viz` tag: https://next.carbondesignsystem.com/assets/components?tags[]=data-visualization

#### Testing / reviewing

The side nav has been reduced to this:
image

There are no changes within the catalogs or any of the libraries (and their assets).


aubrey-oneal (approved):
Looks good. Note that there the homepage links to some hidden pages (like getting started). We could choose to address that in another issue or ignore it for a lower effort.

aagonzales:
Yeah that's a good point about the homepage content @aubrey-oneal. I'm also wondering if we need to add a disclaimer somewhere about why we've removed all the items in the left nav. Maybe we can update the banner to something more informational.

mattrosno:
@aubrey-oneal 

>Hid the home page feature card for how carbon works, the section to start designing and start developing, and the section that links to the roadmap because those pages have been hidden

I hid portions of the home page so there aren't any dead links. Even though it's pretty minimal, there's still enough content on the home page to set context and guide users to the catalogs.

francinelucca (approved):
LGTM!

